PR #12, opened 2016-04-19
Add Undo/Redo fucntions

tiwanari:


tiwanari:
This is an enhancement PR for the app's programming screen.
However, we haven't proved user needs for these functions yet, so please leave it.

PR #22, opened 2016-07-10
Refactor/settings

mandaiy:
yet another pull req for the setting activity.

breaking changes:
- add RxJava/RxAndroid (good bye AsyncTask!)
- layout is changed (two-pane to single-pane with dialog)
- remove "Try to Scan" button in DeviceSelectFragment (formerly called)
  - when the dialog opened, bluetooth device scanning automatically starts (how do you think?)

there are still many hard-coding strings, incoherent naming.
please comment


app/src/main/java/com/pileproject/drive/setting/app/ProgramListFragment.java
```diff
@@ -80,17 +92,25 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
 
+
         return v;
     }
 
     @Override
     public void onActivityCreated(Bundle savedInstanceState) {
         super.onActivityCreated(savedInstanceState);
 
+        final ProgramListFragment thiz = this;
```
mandaiy: seems ugly. any workaround?

mandaiy:
I'd like to discuss
- the behavior of the scanning process (that automatically starts when the user opens the dialog)
- the names of DeviceSelect\* (tentatively BluetoothDevice*)
  with you.
  Any opinions?

app/src/main/java/com/pileproject/drive/setting/app/ProgramListFragment.java
tiwanari: DriveApplication.java can provide a context but it is not good way for this, I think. View v has its context and you can access it with getContext method.

tiwanari:
Question, during search, can users choose a device or go to other page (I mean "can the process stop correctly?")
BluetoothDevice seems good. Is there any problem ?

app/src/main/java/com/pileproject/drive/setting/app/ProgramListFragment.java
mandaiy: It can't because AlertDialogFragment.Builuder needs the parent fragment instance rather than a context in this case.
I think another resort is maintaining 'this' as a field variable, but it gets uglier I think...

mandaiy:
> Question, during search, can users choose a device or go to other page (I mean "can the process stop correctly?")

Of course they can.

> BluetoothDevice seems good. Is there any problem ?

there's no problem, but I need your approval ;)
maybe this naming scheme is suitable, since there could be 'WifiDevice' or something.
or it could be 'BluetoothMachine'. which is preferable?

app/src/main/java/com/pileproject/drive/setting/app/ProgramListFragment.java
tiwanari: Oh, I was confused... I have no good idea now;( what should we do in normal (?) way? Something like this: http://stackoverflow.com/questions/22777928/button-click-inside-fragment-to-open-up-new-fragment ?

tiwanari:
Ok, I also think we should discuss about using "machine" or "device," and in my opinion, machine is better to keep code consistence. 
I have one more question, what should users do when they try to search again?

mandaiy:
> Ok, I also think we should discuss about using "machine" or "device," and in my opinion, machine is better to keep code consistence. 

👍

> I have one more question, what should users do when they try to search again?

Now this app continually scans bluetooth devices (in other words, users do not need to restart it)
To be precise, the scanning restarts when it ends

https://github.com/myusak/drive/blob/30cffb6bb15e0f8add521d8e3645301af4db78c3/app/src/main/java/com/pileproject/drive/setting/machine/BluetoothDeviceSelectFragment.java#L323-L347

When the broadcast receiver receives 'BluetoothAdapter.DISCOVERY_FINISHED', 'onNext' is called.
Then, it calls BluetoothAdapter#startDiscovery() again.

tiwanari:
Perfect :)

app/src/main/java/com/pileproject/drive/setting/app/ProgramListFragment.java
mandaiy: 'ProgramListFragment.this' worked. 
really embarrassed... (and that's why i hate this language)

mandaiy:
I hope you accept my pull request.

tiwanari:
thanks for a notification:) give me time for check.

app/src/androidTest/AndroidManifest.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
```
tiwanari, line 1: Why did you remove this?

app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivityBase.java
```diff
@@ -118,43 +129,37 @@ private void findViews() {
...
-        mSpaceManager.deleteAllBlocks(); // delete existing blocks
-        if (isSample) { // load a sample program
-            mSpaceManager.loadSampleProgram(selectedText);
-        } else { // load a user program
-            mSpaceManager.loadUserProgram(selectedText);
-        }
+    private void moveToTitleActivity() {
+        mSpaceManager.saveExecutionProgram();
+
+        // TODO: go to title activity
```
tiwanari, line 155: Do you plan to keep this as TODO in this PR?

app/src/main/res/layout/fragment_bluetooth_machine_select.xml
```diff
@@ -0,0 +1,108 @@
+<?xml version="1.0" encoding="utf-8"?><!--
```
tiwanari: add a break before "<!--"

tiwanari:
And this PR seems to conflict with the previous PR. Please fix it.

tiwanari:
Though there are some kinds of stuff to be fixed, totally LGTM. APPROVED.

app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivityBase.java
mandaiy, line 155: The app does nothing when this button is pressed even before my pull request.
I didn't come up with a neat way to jump to the title, since the intent differs according
to the concrete title activity.
Can you give me the formal way?

mandaiy:
What are 'some kinds of stuff to be fixed'?
It sounds like the PR contains more issues other than you mentioned.
I'd like to fix these points.

app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivityBase.java
tiwanari, line 155: i understood. so, i think it is good to keep almoslll blank however

mandaiy, line 155: Please fix it on your excellent PR :)

tiwanari:
I'm sorry for confusing you... I mean "there are kinds of some stuff to be fixed (as I mentioned above)". So, all are fixed and let's close this PR. I love it:)

PR #34, opened 2016-09-15
Refactor execution package

tiwanari:
I refactored `execution` package and it affected classes in other packages like `RepetitionBreakBlock`.


app/src/main/java/com/pileproject/drive/execution/ExecutionActivityBase.java
```diff
@@ -310,7 +305,7 @@ private void onConnectionError() {
                 .setCancelable(false)
                 .show();
 
-        // Inform this activity has been disconnected by the device
+        // inform this device has been disconnected by the device
```
mandaiy: Could you please fix this sentence? I could not understand the meaning.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,28 +17,19 @@
...
 import java.util.Stack;
 
 /**
- * This class has the condition of execution
- *
- * @author <a href="mailto:tatsuyaw0c@gmail.com">Tatsuya Iwanari</a>
- * @version 1.0 7-July-2013
+ * A container class that has the condition of program execution.
  */
 public class ExecutionCondition {
```
mandaiy, line 33: I'm not sure what kind of use cases you expect,
but I do not like these access controls since it does not protect itself.
And this seems provide inconsistent interfaces; while ifStack is public, you also add a kind of setter for that.
In short, why not encapsulate this?

mandaiy (commented):
Hi, I had a quick look at your commit.
To my understanding, you haven't made big changes.
Is my understanding right?
If so, basically it's up to you to approve this PR.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
tiwanari: I'm sorry for late replay 😢 

Originally, I prepared `pushSelectionResult(...)` as a helper method. With it, users (mainly the method is used in *Block classes) can use it without considering the program count. But you're right and I will prepare getters and make ifStack private.

tiwanari:
Hi, I fixed codes based on our conversations and simplify the logic of [isThrough()](https://github.com/PileProject/drive/pull/34/files#diff-8a805b762f3f42f7747aa3d3d048eecdR136) method by removing not tested logics for nested ifs. Based on the change, I also changed the appearance of selection blocks like:
screenshot_20160923-013007

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,48 +17,186 @@
...
+        if (block instanceof SelectionEndBlock) return false;
+
+        IfStatus ifStatus = mIfStack.peek();
+        BlockBase nearestSelectionBlock = mBlocks.get(ifStatus.index);
+
+        int middleIf = (nearestSelectionBlock.getRight() + nearestSelectionBlock.getLeft()) / 2;
+        int middle = (block.getRight() + block.getLeft()) / 2;
+
+        // go through this block if (true && left) || (!false && right)
+        return (ifStatus.result && middle <= middleIf) || (!ifStatus.result && middleIf < middle);
```
mandaiy: You should separate the graphics-specific logic from real-programming-related logic, in case you want to revert the block images.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,48 +17,186 @@
...
+     */
+    public void decrementProgramCount() {
+        mProgramCount--;
+    }
+
+    /**
+     * Set the program count.
+     * @param pc
+     */
+    public void setProgramCount(int pc) {
```
mandaiy: does this method really work?

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,48 +17,186 @@
...
-        whileStack = new Stack<>();
-        ifStack = new Stack<>();
-        beginningOfCurrentWhileLoop = -1;
+    private Stack<Integer> mWhileStack;
+    private Stack<IfStatus> mIfStack;
+    private int mBeginningOfCurrentLoop;
+    private int mProgramCount;
+    private ArrayList<BlockBase> mBlocks;
+
+    public ExecutionCondition(ArrayList<BlockBase> blocks) {
```
mandaiy: does this class really depend on the ArrayList? If not so, you should replace with just "List"

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
```diff
@@ -158,155 +117,56 @@ private void finalizeExecution() {
         }
     }
 
-    // Wait on the millisecond time scale
+    // wait in milliseconds
     private void waitMillSec(int milli) {
         try {
-            Thread.sleep(milli); // Sleep for milli sec
+            Thread.sleep(milli); // sleep for milli sec
         } catch (InterruptedException e) {
```
mandaiy, line 114: it would be better to rethrow as non-check exception? same comment applies to other lines

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
```diff
@@ -25,118 +25,78 @@
...
-
-        mCondition.blocks = mManager.loadExecutionProgram();
+        mCondition = new ExecutionCondition(mManager.loadExecutionProgram());
         boolean isStopped = false;
         try {
-            for (mCondition.programCount = 0;
-                 mCondition.programCount < mCondition.blocks.size();
-                 mCondition.programCount++) {
-                // Halt execution
+            for (mCondition.setProgramCount(0);
```
mandaiy: do you have any reasons that you did not include initialize logic of program count in the constructor?
(also it could be better to use while statement? but this heavily depends on your taste)

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,48 +17,186 @@
...
-
-    public ExecutionCondition() {
-        whileStack = new Stack<>();
-        ifStack = new Stack<>();
-        beginningOfCurrentWhileLoop = -1;
+    private Stack<Integer> mWhileStack;
+    private Stack<IfStatus> mIfStack;
+    private int mBeginningOfCurrentLoop;
+    private int mProgramCount;
+    private ArrayList<BlockBase> mBlocks;
```
mandaiy: if the blocks do not change, the field should be final

mandaiy (changes requested):
Totally this PR contains a lot of solid work, particularly the logic becomes really cleaner in programming processing. I would appreciate if you could fix this according to my comments.

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
tiwanari: In this case, InterruptedException often occurs while a program execution to stop the program immediately by a user action (e.g., 'stop button'). So, as for this method, the exception can be ignored. But I will fix other methods, thanks :)

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
tiwanari: Sorry and thanks. It does nothing.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
tiwanari: Finally, I have decided to change it to `List` according to your comments. But let me leave a note for this. The interface of List does not specify the time complexity of `get(int index)` and this program needs random accesses frequently. That's why I have used ArrayList not other List. But the complexity can be ignored because the size of the list is small enough in practical.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
tiwanari: Thanks and what do you think where we should keep it? ExecutionThread was the original location of the long version of this codes.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
mandaiy: Sorry, I missed the use of <code>get</code>. In my opinion you should stick to <code>ArrayList</code>. Or you could convert <code>List</code> into  <code>ArrayList</code> if the list is not <code>instanceof RandomAccess</code>

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
mandaiy: The best way I came up with is to make a new class that holds graphics-specific-if logic, which might implement an interface common with other strategies.
Or the logic could be in <code>Block</code> class?

I think in future we should separate geometric information from <code>BlockBase</code> class because we are going to implement text-based programming interface in our app.
We can use one <code>BlockBase</code> instance as an atomic operation, which does not need geometric information. When users write some code in our app, then we should parse it, convert it into list of <code>BlockBase</code> instances then we can reuse our code base we already have (though we might need more pretty class name).

So the geometric information should leave from the block class and a new interface, say <code>VisualBlock</code> interface or whatever, should be responsible for the information.
Then the if-logic should be in the interface.

But if we could compile a program and could know next indexes of if-blocks, this conversation would be useless.

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
mandaiy: that's why I hate Java and checked-exceptions.

tiwanari:
I fixed codes based on our discussion and added a logic class for visual block programming.

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
```diff
@@ -146,167 +105,62 @@ public void run() {
...
 
-    // Wait on the millisecond time scale
-    private void waitMillSec(int milli) {
+    // wait in milliseconds
+    private void waitMilliSec(int milli) {
         try {
-            Thread.sleep(milli); // Sleep for milli sec
+            Thread.sleep(milli); // sleep for milli sec
         } catch (InterruptedException e) {
             e.printStackTrace();
```
mandaiy: Is this <code>e.printStackTrace()</code> necessary?

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,48 +17,196 @@
...
+        return mProgramCount >= mBlocks.size();
+    }
+
+    /**
+     * Get the specified block with an index.
+     * @param index
+     * @return
+     * @throws IndexOutOfBoundsException
+     */
+    public BlockBase getBlock(int index) throws IndexOutOfBoundsException {
```
mandaiy: Since users of this class should not have access to blocks by indexing, it's better to hide this method by adding a getter with a selection result instance.

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -17,48 +17,196 @@
...
+    public void decrementProgramCount() {
+        mProgramCount--;
+    }
+
+    /**
+     * Set the program count.
+     * @param pc program count
+     * @throws IndexOutOfBoundsException
+     */
+    public void setProgramCount(int pc) throws IndexOutOfBoundsException {
```
mandaiy: Stray method; It can be removed. Or do you have particular use cases for this?

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
```diff
@@ -25,119 +25,78 @@
...
+            while (!mCondition.hasProgramFinished()) {
+
+                // halt execution
                 if (mHalt) {
                     break;
                 }
 
-                // Stop temporarily
+                // stop temporarily
                 if (mStop) {
```
mandaiy: It's really painful to understand this if block.
If <code>mController.halt()</code> is idempotent, you could simplify this?

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
```diff
@@ -146,167 +105,62 @@ public void run() {
...
-    private void finalizeExecution() {
+    private void finalizeExecution() throws RuntimeException {
         mHalt = true;
         mStop = false;
-        mThreadNum = 0;
 
         try {
             mController.finalize();
         } catch (RuntimeException e) {
             e.printStackTrace();
```
mandaiy: Do you really need this catch clause and <code>e.printStackTrace()</code>?

mandaiy (changes requested):
Added some notes. Some of them are bit controversial, so I'm open to a discussion.

app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java
tiwanari: > If mController.halt() is idempotent, you could simplify this?

Yeah, this is true. But `halt()` always sends `stop` commands (\* 2 for left&right) to a robot and it floods buffers. Anyway, I'll try to make it clearer.

tiwanari:
Updated codes for check.

mandaiy (approved):

Issue #36, opened 2016-09-24
Enable TravisCI


makotoshimazu:

tiwanari:
Hi @amiq11,

This is done on #37. So I close this issue. Thanks!

PR #37, opened 2016-09-24
Added .travis.yml

makotoshimazu:
This patch enables the bot to test this repository automatically.

Issue #36


tiwanari:
I'm tackling the `cannot find symbol` problem. In a local machine, the error does not occur with `./gradlew build connectedCheck` but I met other error `Could not launch activity` after the building process causing `cannot find symbol` problem so I fixed it at first. Since the error also made me confused, let me a note for that.

The error has occurred in `com.pileproject.drive.util.fragment.AlertDialogFragmentTest#testCheckComponent_titleString` because it could't find  `com.pileproject.drive.util.fragment.TestActivity`. But the Activity class was surely declared in `androidTest/AndroidManifest.xml` and should have been found in test.

Finally, I came across [a discussion on StackOverFlow](http://stackoverflow.com/questions/26244998/androidmanifest-in-androidtest-directory-being-ignored) and we should prepare AndroidManifest in `debug` folder to follow [the default build type](https://developer.android.com/studio/test/index.html#change_the_test_build_type) while testing.

[UPDATE]
As follow the instruction, I changed a `sourceSets` element from `test` to `debug` and this solved the `Could not launch activity` error.

makotoshimazu:
#44 covers this. closed.

PR #39, opened 2016-09-25
Remove a useless conversion

tiwanari:
Sorry, I want to revert a stupid code change on #34

### DIFF
--- a/app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
+++ b/app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
@@ -48,9 +48,6 @@ public class ExecutionCondition {
     }
 
     public ExecutionCondition(ArrayList<BlockBase> blocks) {
-        // convert a list with random access support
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
```diff
@@ -48,9 +48,6 @@ public SelectionResult(int index, boolean result) {
     }
 
     public ExecutionCondition(ArrayList<BlockBase> blocks) {
```
mandaiy: Hmm, your choice was remove. That's ok, but you could change the signature from ArrayList to List.

mandaiy (approved):

app/src/main/java/com/pileproject/drive/execution/ExecutionCondition.java
tiwanari: Thanks. The signature may confuse and I change it.

PR #42, opened 2016-10-05
Fix formatting error in Unit.java

mandaiy:
This commit resolves an error which occurred
at the moment of generating an NTimes block.
The cause of the bug was casting illegally
from BigDecimal to Integer class in Unit.java.


tiwanari (approved):
LGTM! :+1:

tiwanari:
I think you haven't noticed this PR is approved, @myusak :)

mandaiy:
Thanks for your headsup

PR #44, opened 2016-10-05
Enable TravisCI

mandaiy:
## Summary

This pull-request introduces continuous integration on TravisCI.
This also resolves and closes #37.

This PR also removes the test in androidTest which uses Activity.
The test failed due to the package problem, which should be resolved in future.
The cause of the failure was Gradle cannot determine correct packages during testing,
as @tiwanari mentioned in #37.
(The test seemed awkward compared to ones other developers created.
So anyway the test had to be modified.)

Finally the failure of NumberTextViewDelegate test was resolved, which was caused by
the writer's fault and lack of understanding.
## Description of the Failure

As you can see [here](https://travis-ci.org/myusak/drive/branches), 
while develop-build#18 failed, develop-build#19 succeeded, 
the only difference is the JDK version in .travis.yml.
So I should conclude the cause was the version, unbelievably and unfortunately.
## So...

Should we stick to `oraclejdk7`? or just leave it?
Give me your thoughts (but I'm really fed up with it, please take it over).

And I know we need more snazzy `.travis.yml`. Please edit it.


tiwanari:
Oh, you finally made it! 👍

The following is random thoughts and comments.

### Java version

Version problems always keep us upset, doesn't it? 😫  I don't think we have to stick to `oraclejdk7` because there are no concrete reasons not to use 8. I just specified the version based on a local computer environment.

### Activity package problem

As for the package problem of Activity, I found a solution and updated #37 a little. Please try to change the build type [here](https://github.com/myusak/drive/blob/test/travis-ci/app/build.gradle#L64) from `test` to `debug` like [this](https://github.com/PileProject/drive/blob/e59acb3477a54816462dca221c8e81198caae26b/app/build.gradle#L65). This will enable `gradle` to find Activities but I'm not sure all tests pass after the change and agree with you that we should modify the test.

### Test command

I'm sorry I forgot doing some experiments on #37 and informing you which are experiments (or core changes). According to [this page](https://docs.travis-ci.com/user/languages/android), the default test command is 

```
./gradlew build connectedCheck
```

I think this is better because `build` does tests and `connectedCheck` all device checks in my understanding. Sorry again.

makotoshimazu:
For java version, I guess we should define that based on the available version on Android, right?
If I understand correctly, we should use JDK7 because our application is targeting on Android I or later.
WDYT?

makotoshimazu:
Sorry, I feel the lack of my explanation.
I wish deployment would be also done by travis, so I think we should be care for the JDK version for building the apk.

mandaiy:
> Please try to change the build type here from test to debug like this.

Thank you but as I mentioned, I feel the test has to be changed.
For this PR I'd like to concentrate on the setting of Travis.
Could you hold on and wait for next my PR?

> I wish deployment would be also done by travis, so I think we should be care for the JDK version for building the apk.

To my understanding, jdk8 is able to compile java7 or below code and produce byte code with java7 compatibility, by adding `targetCompatibility` in Gradle or `javac -target 1.7` (is it correct?).
And for my environment, I'm using jdk8 (`"1.8.0_101"`) and can produce APKs for sdk15 or later (which is Android I).
So technically we are using jdk8, same as the current TravisCI.

makotoshimazu:
LGTM

Please wait for tiwanari@ 's review.

tiwanari:
> Could you hold on and wait for next my PR?

Sure! I think so, too :)

btw, why not use `build` with `./gradlew` at the `script` section? Is there any error?

tiwanari (approved):
Totally, LGTM! Thanks!

mandaiy:
`build` is `assemble` + `check`.
`assemble` is a task for creating jar file, which does not make sense in CI server.
http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Build-Tasks

tiwanari:
> build is assemble + check.

I see. Thanks for the reference :)

PR #46, opened 2016-10-13
Hotfix/execution bugs

tiwanari:
I checked #3 and fortunately the issue was (almost) solved 🎉 

However, unfortunately, there are many bugs in some classes: 
- [1st commit](https://goo.gl/K5RXEH)
  - `ExecutionThread`: The Program Count hadn't been updated in a while loop.
  - `BlockProgramLogic`: Some return clauses returned opposite values.
- [2nd commit](https://goo.gl/57fqlw)
  - `BluetoothCommunicator`: This is not a bug but a notice was lacking and this made it difficult to know whether the second exception occurred or not (related to the 4th commit below).
- [3rd commit](https://goo.gl/kafp8v)
  - `ExecutionActivityBase`: A request code setting method hadn't been called and thus the class didn't finish after executions.
- [4th commit](https://goo.gl/T1iB4j)
  - Remove `is_connected` flag that was uglily scatted in many places. It was introduced to avoid useless reconnection to a device but some code changes, especially around `MachineController`, had conflicted the mechanism and caused connection errors. So, I forced `ExecutionActivityBase` to `connect` and `disconnect` (this was lacking in the old code) before / after execution respectively.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -71,19 +71,22 @@ dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
 
     testCompile 'junit:junit:4.12'
+    // mockito
+    testCompile 'org.mockito:mockito-core:1.10.19'
 
+    // android test libraries
     androidTestCompile 'com.android.support.test:runner:0.5'
     androidTestCompile 'com.android.support.test.espresso:espresso-core:2.2.2'
-
-    // specify the version of this library
-    // (without this line gradle sync fails due to the version conflict)
-    androidTestCompile 'com.android.support:support-annotations:23.4.0'
 
     // mockito
     androidTestCompile 'org.mockito:mockito-core:1.10.19'
     androidTestCompile "com.crittercism.dexmaker:dexmaker:1.4"
     androidTestCompile "com.crittercism.dexmaker:dexmaker-dx:1.4"
     androidTestCompile "com.crittercism.dexmaker:dexmaker-mockito:1.4"
+
+    // specify the version of this library
+    // (without this line gradle sync fails due to the version conflict)
+    androidTestCompile 'com.android.support:support-annotations:23.4.0'
 
     // our library
     compile 'com.pileproject:drivecommand:2.1.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -67,9 +67,7 @@ android {
     }
 }
 
-apply plugin: 'maven'
-def dexmakerVersion = '17cc9cc'
-def dexmakerGroupId = 'com.github.crittercism.dexmaker'
+def dexMakerVersion = '1.4'
 
 dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
@@ -84,12 +82,9 @@ dependencies {
 
     // mockito
     androidTestCompile 'org.mockito:mockito-core:1.10.19'
-    androidTestCompile "${dexmakerGroupId}:dexmaker:${dexmakerVersion}"
-    androidTestCompile "${dexmakerGroupId}:dexmaker-dx:${dexmakerVersion}"
-    androidTestCompile "${dexmakerGroupId}:dexmaker-mockito:${dexmakerVersion}"
-    // androidTestCompile "com.crittercism.dexmaker:dexmaker:1.4"
-    // androidTestCompile "com.crittercism.dexmaker:dexmaker-dx:1.4"
-    // androidTestCompile "com.crittercism.dexmaker:dexmaker-mockito:1.4"
+    androidTestCompile "com.crittercism.dexmaker:dexmaker:${dexMakerVersion}"
+    androidTestCompile "com.crittercism.dexmaker:dexmaker-dx:${dexMakerVersion}"
+    androidTestCompile "com.crittercism.dexmaker:dexmaker-mockito:${dexMakerVersion}"
 
     // specify the version of this library
     // (without this line gradle sync fails due to the version conflict)
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -67,6 +67,10 @@ android {
     }
 }
 
+apply plugin: 'maven'
+def dexmakerVersion = '17cc9cc'
+def dexmakerGroupId = 'com.github.crittercism.dexmaker'
+
 dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
 
@@ -80,9 +84,12 @@ dependencies {
 
     // mockito
     androidTestCompile 'org.mockito:mockito-core:1.10.19'
-    androidTestCompile "com.crittercism.dexmaker:dexmaker:1.4"
-    androidTestCompile "com.crittercism.dexmaker:dexmaker-dx:1.4"
-    androidTestCompile "com.crittercism.dexmaker:dexmaker-mockito:1.4"
+    androidTestCompile "${dexmakerGroupId}:dexmaker:${dexmakerVersion}"
+    androidTestCompile "${dexmakerGroupId}:dexmaker-dx:${dexmakerVersion}"
+    androidTestCompile "${dexmakerGroupId}:dexmaker-mockito:${dexmakerVersion}"
+    // androidTestCompile "com.crittercism.dexmaker:dexmaker:1.4"
+    // androidTestCompile "com.crittercism.dexmaker:dexmaker-dx:1.4"
+    // androidTestCompile "com.crittercism.dexmaker:dexmaker-mockito:1.4"
 
     // specify the version of this library
     // (without this line gradle sync fails due to the version conflict)
--- a/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
+++ b/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
@@ -20,25 +20,14 @@ import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.internal.util.reflection.Whitebox;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
 
 @RunWith(AndroidJUnit4.class)
 public class BluetoothCommunicatorTest {
 
-    // @Test(expected = NullPointerException.class)
-    // public void whenBluetoothDeviceIsNull_thenThrowNullPointerException() throws Exception {
-    //     // @NonNull works properly
-    //     // it raises a warning
-    //     new BluetoothCommunicator(null);
-    // }
+    @Test(expected = NullPointerException.class)
+    public void whenBluetoothDeviceIsNull_thenThrowNullPointerException() throws Exception {
+        new BluetoothCommunicator(null);
+    }
 
     @Test
     public void whenFirstOpenMethodFails_thenTrySecondMethod_thenThrowIOException() throws Exception {
@@ -56,90 +45,99 @@ public class BluetoothCommunicatorTest {
 
     @Test
     public void whenSocketIsAlreadyNull_thenCloseSucceeds() throws Exception {
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        comm.close();
+        // comm.close();
     }
 
     @Test
     public void whenOutputIsNull_thenWritesNothing() throws Exception {
-        OutputStream outputStream = mock(OutputStream.class);
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // OutputStream outputStream = mock(OutputStream.class);
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        // use a mock instance
-        Whitebox.setInternalState(comm, "mOutputStream", outputStream);
+        // // use a mock instance
+        // Whitebox.setInternalState(comm, "mOutputStream", outputStream);
 
-        // write null
-        comm.write(null, 0);
-        verify(outputStream).write(null);
+        // // write null
+        // comm.write(null, 0);
+        // verify(outputStream).write(null);
     }
 
     @Test
     public void whenOutputIsArray_thenWritesTheArray() throws Exception {
-        OutputStream outputStream = mock(OutputStream.class);
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // OutputStream outputStream = mock(OutputStream.class);
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        // use a mock instance
-        Whitebox.setInternalState(comm, "mOutputStream", outputStream);
+        // // use a mock instance
+        // Whitebox.setInternalState(comm, "mOutputStream", outputStream);
 
-        // write a data
-        byte[] data = {0x00, 0x01, 0x02, };
-        comm.write(data, 0);
-        verify(outputStream).write(data);
+        // // write a data
+        // byte[] data = {0x00, 0x01, 0x02, };
+        // comm.write(data, 0);
+        // verify(outputStream).write(data);
     }
 
-    @Test(expected = RuntimeException.class)
+    // @Test(expected = RuntimeException.class)
+    @Test
     public void whenWriteFails_thenThrowRuntimeException() throws Exception {
-        OutputStream outputStream = mock(OutputStream.class);
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // OutputStream outputStream = mock(OutputStream.class);
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        // write a data
-        byte[] data = {0x00, 0x01, 0x02, };
+        // // write a data
+        // byte[] data = {0x00, 0x01, 0x02, };
 
-        // use a mock instance
-        Whitebox.setInternalState(comm, "mOutputStream", outputStream);
-        doThrow(IOException.class).when(outputStream).write(data);
+        // // use a mock instance
+        // Whitebox.setInternalState(comm, "mOutputStream", outputStream);
+        // doThrow(IOException.class).when(outputStream).write(data);
 
-        comm.write(data, 0);
+        // comm.write(data, 0);
     }
 
     @Test
     public void whenInputIsNull_thenReadNothing() throws Exception {
-        InputStream inputStream = mock(InputStream.class);
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // InputStream inputStream = mock(InputStream.class);
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        // use a mock instance
-        Whitebox.setInternalState(comm, "mInputStream", inputStream);
+        // // use a mock instance
+        // Whitebox.setInternalState(comm, "mInputStream", inputStream);
 
-        // read null
-        comm.read(0, 0);
-        verify(inputStream).read(new byte[0]);
+        // // read null
+        // comm.read(0, 0);
+        // verify(inputStream).read(new byte[0]);
     }
+
     @Test
-
     public void whenInputIsArray_thenReadTheArray() throws Exception {
-        InputStream inputStream = mock(InputStream.class);
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // InputStream inputStream = mock(InputStream.class);
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        // use a mock instance
-        Whitebox.setInternalState(comm, "mInputStream", inputStream);
+        // // use a mock instance
+        // Whitebox.setInternalState(comm, "mInputStream", inputStream);
 
-        // read a data
-        comm.read(100, 0);
-        verify(inputStream).read(new byte[100]);
+        // // read a data
+        // comm.read(100, 0);
+        // verify(inputStream).read(new byte[100]);
     }
 
-    @Test(expected = RuntimeException.class)
+    // @Test(expected = RuntimeException.class)
+    @Test
     public void whenReadFails_thenThrowRuntimeException() throws Exception {
-        InputStream inputStream = mock(InputStream.class);
-        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+        // TODO: mock BluetoothDevice
+        // InputStream inputStream = mock(InputStream.class);
+        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
 
-        // use a mock instance
-        Whitebox.setInternalState(comm, "mInputStream", inputStream);
-        doThrow(IOException.class).when(inputStream).read(new byte[100]);
+        // // use a mock instance
+        // Whitebox.setInternalState(comm, "mInputStream", inputStream);
+        // doThrow(IOException.class).when(inputStream).read(new byte[100]);
 
-        // read a data
-        comm.read(100, 0);
-        verify(inputStream).read(new byte[100]);
+        // // read a data
+        // comm.read(100, 0);
+        // verify(inputStream).read(new byte[100]);
     }
 }
--- a/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
+++ b/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
@@ -16,7 +16,6 @@
 
 package com.pileproject.drive.comm;
 
-import android.bluetooth.BluetoothDevice;
 import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.Test;
@@ -34,10 +33,12 @@ import static org.mockito.Mockito.verify;
 @RunWith(AndroidJUnit4.class)
 public class BluetoothCommunicatorTest {
 
-    @Test(expected = IOException.class)
-    public void whenBluetoothDeviceIsNull_thenThrowIOException() throws Exception {
-        BluetoothDevice device = null;
-        BluetoothCommunicator comm = new BluetoothCommunicator(device);
+    @Test
+    // @Test(expected = NullPointerException.class)
+    public void whenBluetoothDeviceIsNull_thenThrowNullPointerException() throws Exception {
+        // @NonNull works properly
+        // it causes a warning
+        new BluetoothCommunicator(null);
     }
 
     @Test
@@ -57,6 +58,7 @@ public class BluetoothCommunicatorTest {
     @Test
     public void whenSocketIsAlreadyNull_thenCloseSucceeds() throws Exception {
         BluetoothCommunicator comm = new BluetoothCommunicator(null);
+
         comm.close();
     }
 
@@ -112,7 +114,7 @@ public class BluetoothCommunicatorTest {
 
         // read null
         comm.read(0, 0);
-        verify(inputStream).read(null);
+        verify(inputStream).read(new byte[0]);
     }
     @Test
 
--- a/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
+++ b/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
@@ -29,115 +29,115 @@ public class BluetoothCommunicatorTest {
         new BluetoothCommunicator(null);
     }
 
-    @Test
-    public void whenFirstOpenMethodFails_thenTrySecondMethod_thenThrowIOException() throws Exception {
-        // TODO: write a good test
-        // NOTE: can't mock/spy BluetoothDevice/BluetoothSocket
-        //       because they are `final` classes
-    }
+    // TODO: consider using PowerMock
 
-    @Test
-    public void whenSocketIsNotNull_thenCloseSucceeds() throws Exception {
-        // TODO: write a good test
-        // NOTE: can't mock/spy BluetoothSocket
-        //       because it is a `final` class
-    }
+    // @Test
+    // public void whenFirstOpenMethodFails_thenTrySecondMethod_thenThrowIOException() throws Exception {
+    //     // TODO: write a good test
+    //     // NOTE: can't mock/spy BluetoothDevice/BluetoothSocket
+    //     //       because they are `final` classes
+    // }
 
-    @Test
-    public void whenSocketIsAlreadyNull_thenCloseSucceeds() throws Exception {
-        // TODO: mock BluetoothDevice
-        // BluetoothCommunicator comm = new BluetoothCommunicator(null);
+    // @Test
+    // public void whenSocketIsNotNull_thenCloseSucceeds() throws Exception {
+    //     // TODO: write a good test
+    //     // NOTE: can't mock/spy BluetoothSocket
+    //     //       because it is a `final` class
+    // }
 
-        // comm.close();
-    }
+    // @Test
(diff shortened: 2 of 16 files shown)


tiwanari:
Sorry, I forgot adding null checks and to commit it (yeah, actually they are related to #3).

mandaiy:
I did not check thoroughly your changes yet, but could you give here how you have tested your changes? 
I guess you did on a real robot and your tablet but in case you tested something other than on them.

And from my quick look, it seems easier to write test code for `BlockProgrammingLogic`.
For `NxtController` if you remove the logic related to preferences, the class might be a test-friendly class.

tiwanari:
Thanks for your comments :)
I'll add more detail. Can you give me a moment?

tiwanari:
Let me leave a note. Actually, as you said, most of the tests are running some programs. If you expected some test programs, to tell the truth, there were no concrete tests for them.

Firstly, I executed a blank program. At this point, my NXT hasn't sensors nor motors and the program successfully finished. This meant `NxtControllerBuilder` has finished his life as for motors because `ExecutionThread` makes commands to NXT motors after execution even if the program was blank. This lacks tests for sensors and I noticed `NxtController` has no null checks (this is the latest commit). With the change, #3 were surely solved.

However, `ExecutionActivity` didn't finish after execution, i.e., I couldn't go back to `ProgrammingActivity`. So, I dug into some classes such as `ExecutionCondition` because there is a possibility that the program was caught in a forever loop and, of course,  I checked `ExecutionActivityBase`.  I found most of the bugs manually in this stage.

The 4th commit is totally different from others. When I tried to run a program twice or more after fixing other bugs, I always caught a connection error. Tracing an execution, I found `NxtMachine (mMachine)` in `NxtExecutionActivity` hadn't been initialized and it caused an Exception. The lack of initialization was caused by the replacement of our old library with new one `drivecommand` and I decided to remove all uglily `is_connected` flags.

tiwanari:
So, please give me a little time to add tests as you advised :)

tiwanari:
Hi, I added some tests and fixed bugs.

BTW, how about using [PowerMock](https://goo.gl/iBmuW2)? I have trouble testing `final` classes because they cannot be mocked/spied with Mockito. 

Note: I don't like `Whitebox.foo()` in some test codes which is a part of Mockito ...

app/build.gradle
```diff
@@ -71,20 +71,23 @@ dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
 
     testCompile 'junit:junit:4.12'
+    // mockito
+    testCompile 'org.mockito:mockito-core:1.10.19'
```
mandaiy, line 77: Just a question: is this needed? `androidTestCompile` has the same repository.

app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+@RunWith(AndroidJUnit4.class)
+public class BluetoothCommunicatorTest {
+
+    @Test(expected = IOException.class)
+    public void whenBluetoothDeviceIsNull_thenThrowAnIOException() throws Exception {
+        BluetoothDevice device = null;
+        BluetoothCommunicator comm = new BluetoothCommunicator(device);
```
mandaiy: Well, it's not your changes but should `BluetoothComunicator` accept `null`?
I think it's better to check beforehand and throw an exception if the argument is `null`

app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void testWrite() throws Exception {
+        OutputStream outputStream = mock(OutputStream.class);
+        BluetoothCommunicator comm = new BluetoothCommunicator(null);
+
+        // use a mock instance
+        Whitebox.setInternalState(comm, "mOutputStream", outputStream);
+
+        // write null
+        comm.write(null, 0);
+        verify(outputStream).write(null);
```
mandaiy: better to split these lines into two test methods.
One is `whenInputIsNull_thenWritesNothing` and the other is `whenInputIsArray_thenWritesTheArray` or something.
And you might become happier if you add a sad-path.

Same comment applies `testRead`

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,195 @@
...
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+@RunWith(AndroidJUnit4.class)
+public class BlockProgramLogicTest {
+
+    @Mock ExecutionCondition.SelectionResult trueResult;
```
mandaiy: I think you should not use `@Mock` for `SelectionResult` because it is a **value object**.
Just instantiate it.

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,195 @@
...
+        Whitebox.setInternalState(trueBlock, "mRight", 99);
+
+        // doReturn(100).when(falseBlock).getLeft();
+        // doReturn(199).when(falseBlock).getRight();
+        Whitebox.setInternalState(falseBlock, "mLeft", 100);
+        Whitebox.setInternalState(falseBlock, "mRight", 199);
+    }
+
+    @Test
+    public void whenExecuteConditionHasNoSelectionResult_thenExecuteCurrentBlock() throws Exception {
```
mandaiy: `thenReturnsTrue` is more suitable for the test case, I think.

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+        ExecutionCondition condition = mock(ExecutionCondition.class);
+        doReturn(0).when(condition).sizeOfSelectionResult(); // no selection result
+
+
+        assertTrue(BlockProgramLogic.willCurrentBlockBeExecuted(condition));
+
+
+        verify(condition).sizeOfSelectionResult();  // do call
```
mandaiy: I feel these `verify`s are kind of over-kill.
What we want to ensure is just "if condition has no condition then returns true".
We should not go into implementation too much.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+    SoundSensor soundSensor;
+    LineSensor lineSensor;
+
+    Motor leftMotor;
+    Motor rightMotor;
+
+    @Before
+    public void setUp() {
+        machine = mock(NxtMachine.class);
+        controller = new NxtController(machine);
```
mandaiy, line 57: Maybe you can use `@InjectMocks` annotation.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        soundSensor = mock(SoundSensor.class);
+        lineSensor = mock(LineSensor.class);
+
+        leftMotor = mock(Motor.class);
+        rightMotor = mock(Motor.class);
+    }
+
+    private void setUpMotors() {
+        Whitebox.setInternalState(controller, "mLeftMotor", leftMotor);
+        doCallRealMethod().when(leftMotor).setSpeed(anyInt());
```
mandaiy: What the heck is this?
Why do you need `doCallRealMethod` instead of `verify`?
See my comment in `testSetMotorPower`

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        assertEquals(0, leftMotor.getSpeed());
+        assertEquals(0, rightMotor.getSpeed());
+
+        controller.moveForward();
+
+        // after a move command, they will be initialized with the default values
+        assertEquals(NxtController.INIT_MOTOR_POWER, leftMotor.getSpeed());
+        assertEquals(NxtController.INIT_MOTOR_POWER, rightMotor.getSpeed());
+
+        controller.setMotorPower(NxtController.MotorKind.LeftMotor, 100);
```
mandaiy: It's better to split into two methods:
`whenMotorSpeedsAreNotInitialized_thenMovesForwardWithDefaultValue` and `whenSetMotorPowerCalled_thenMovesForwardWithTheValue`

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        controller.setMotorPower(NxtController.MotorKind.RightMotor, 100);
+
+        controller.moveForward(); // same as above
+
+        assertEquals(NxtController.MAX_MOTOR_POWER, leftMotor.getSpeed());
+        assertEquals(NxtController.MAX_MOTOR_POWER, rightMotor.getSpeed());
+    }
+
+    @Test
+    public void testFinalize() throws Exception {
```
mandaiy: You can rename it to `whenControllerFinalizes_thenMachineDisconnects`

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        doCallRealMethod().when(leftMotor).setSpeed(anyInt());
+        doCallRealMethod().when(leftMotor).getSpeed();
+
+        Whitebox.setInternalState(controller, "mRightMotor", rightMotor);
+        doCallRealMethod().when(rightMotor).setSpeed(anyInt());
+        doCallRealMethod().when(rightMotor).getSpeed();
+    }
+
+
+    // TODO: add tests related to preferences
```
mandaiy, line 73: We leaved `Builder`, but why not reconsider it?
Maybe in this time the name is not `Builder`, but something like `IOProvider`.
The interface should have `getOutputPortA`, `getInputPort1`, ...
And one of the concrete classes is `PreferenceIOProvider`.
With that kind of class you can easily mock it and this file can be moved to `test` directory not `androidTest` directory.

app/src/nxt/java/com/pileproject/drive/execution/NxtController.java
```diff
@@ -208,17 +211,18 @@ public void setMotorPower(MotorKind kind, double percent) {
 
         switch (kind) {
             case LeftMotor:
-                mLeftMotorPower = (int) ((MAX_MOTOR_POWER * percent) / 100.0);
+                mLeftMotorPower = (int) percent;
```
mandaiy: should we change the type of `percent` to `int`?

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        verify(leftMotor).stop();
+        verify(rightMotor).stop();
+    }
+
+    @Test
+    public void testSetMotorPower() throws Exception {
+        setUpMotors();
+
+        // at first, they have not been initialized
```
mandaiy: These two lines are meaningless as this test is for the controller and in general I believe
**states** should not be tested, but **behaviors**.

mandaiy (changes requested):
I do not disagree with introducing Powermock.
Particularly for `RelativeLayout`s.
But I do not know it's good idea to mocking `BluetoothDevice` in unit tests.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,214 @@
...
+        assertEquals(0, leftMotor.getSpeed());
+        assertEquals(0, rightMotor.getSpeed());
+
+        controller.moveForward();
```
mandaiy, line 196: You do not have to `assertEquals` motors' speed. Just `verify` them

``` java
controller.moveForward();
verify(motor).setSpeed(INIT_MOTOR_POWER);
```

app/build.gradle
tiwanari: I think this is necessary to use mockito to classes in `test`.

app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
tiwanari: Sounds good 👍  I will do.

app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java
tiwanari: Will do :)

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
tiwanari: Will do :)

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
tiwanari: Sounds good 👍

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
tiwanari: Yeah, I agree.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: To tell the truth, it causes an Exception maybe because of some issues like https://github.com/mockito/mockito/issues/243 . This can be solved by upgrading mockito to 2.x (I checked) but it conflicts some changes here such as `Whitebox`. To solve them, we should consider alternative ways to test private class/method/member.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: Thanks. I understood 👍

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: Yeah, I agree with the idea and I also considered it but I concluded doing it in this PR is too much. Let's make another PR for this.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: Will do :)

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: Will do :)

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: Will do, thanks :)

app/src/nxt/java/com/pileproject/drive/execution/NxtController.java
tiwanari: `percent` is a misleading name. They are used as `speed` actually. I will fix the type/name of them. Thanks :)

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
tiwanari: Got it :)

tiwanari:
Thanks for comments, @myusak :)

So, I'll try to introduce PowerMock and make some more tests 💪

tiwanari:
Hmm... I have trouble incorporating PowerMock because of a bug of DexMaker, poor compatibleness between PowerMock and DexMaker and so on :cyclone:

In short, using PowerMock along with DexMaker is impossible at present because of incompatibility ([How to get Powermock to work with Dexmaker - Stack Overflow](https://goo.gl/kagICc)). Therefore, we cannot use PowerMock in `androidTest`.

tiwanari:
Hi @myusak,

I fixed codes based on your advice, however, I couldn't use PowerMock at present because of the problem I mentioned above. This does not permit me to remove `Whitebox` and throw an Exception when null is passed to `BluetoothCommunicator` in order to do other tests because `BluetoothDevice` cannot be mocked and I should pass null to it in other tests. 

At this point, I use @NonNull annotation as a workaround (compromise) but we may as well introduce a wrapper such as `BluetoothDeviceProvider` that can be mocked easily and thus enables testing `BluetoothCommunicator`. 

What do you think? Should we do that in this PR? If so, I will also implement such class for `NxtController` as you pointed out.

mandaiy:
I believe that actual code comes first before test code.
I think the `BluetoothComunicator` should reject null object in its constructor and
for now you may as well skip the test code.
When the problem of Powermock is solved, we'd go for that.

I cannot agree with introducing such wrapper interface, since users of the communicator don't
want to be bothered by that.
Feel free to criticise me if my understanding is wrong.

tiwanari:
Thanks for your comments :)
I agreed and commented out all tests related to PowerMock.
So, please review codes again 🙇

tiwanari (commented):
Oh, sorry. I forgot to leave a note for [this commit](https://github.com/PileProject/drive/pull/46/commits/6fe02053e19a0750abd53c05c28268a4d5a6abbc).
This is necessary for `MockitoAnnotations.initMocks()` in tests because there is a bug in DexMaker 1.4. It was fixed at [`17cc9cc`](https://github.com/crittercism/dexmaker/commit/17cc9ccdd52a8b27793ec12f3cebb93ddf469af4) but there is no new release because DexMaker is kind of dead... 
I followed [this instruction](https://github.com/crittercism/dexmaker/issues/32).

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,206 @@
...
+        Whitebox.setInternalState(controller, "mRightMotor", rightMotor);
+    }
+
+
+    // TODO: add tests related to preferences
+
+
+    @Test
+    public void whenTouchSensorIsNull_thenReturnFalse() throws Exception {
+        assertFalse(controller.getTouchSensorValue());
```
mandaiy, line 80: I think this method become more informative if you add `WhiteBox.setInternalState(controller, "mTouchSensor", null)` .

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,206 @@
...
+        Whitebox.setInternalState(controller, "mSoundSensor", soundSensor);
+
+        doReturn(10).when(soundSensor).getDb();
+        assertEquals(10, controller.getSoundSensorValue());
+
+        doReturn(100).when(soundSensor).getDb();
+        assertEquals(100, controller.getSoundSensorValue());
+
+        doReturn(-10).when(soundSensor).getDb();
+        assertEquals(-10, controller.getSoundSensorValue());
```
mandaiy: Why do we need so many asserts?
These don't seem boundary values, so I think it's enough to check on just one arbitrary value.
(Do we need to check on all values in the domain of `int`?)

I also think you should write test-methods for one particular objective. (i.e., do not test two or more things in a method)

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        doReturn(200).when(lineSensor).getSensorValue();
+        assertEquals(200, controller.getLineSensorValue());
+
+        doReturn(-20).when(lineSensor).getSensorValue();
+        assertEquals(-20, controller.getLineSensorValue());
+    }
+
+    @Test
+    public void testMoveForward() throws Exception {
```
mandaiy: Generally speaking it's good practice to name methods in a same format if possible and here it seems possible.
I'd like to suggest `whenMachineMovesForward_thenMotorsMoveCorrectly`

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
```diff
@@ -18,6 +18,7 @@
 
 import android.bluetooth.BluetoothDevice;
 import android.bluetooth.BluetoothSocket;
+import android.support.annotation.NonNull;
 
 import com.pileproject.drivecommand.model.com.ICommunicator;
```
mandaiy, line 24: You might become happier with [Objects#requireNonNull](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.util.function.Supplier-)

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
```diff
@@ -39,31 +39,30 @@
     private OutputStream mOutputStream;
     private InputStream mInputStream;
 
-    public BluetoothCommunicator(BluetoothDevice device) {
+    public BluetoothCommunicator(BluetoothDevice device) throws NullPointerException {
+        if (device == null) {
+            throw new NullPointerException("Device should not be null");
```
mandaiy, line 45: You might become happier with [Objects#requireNonNull](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.util.function.Supplier-)

app/src/nxt/java/com/pileproject/drive/execution/NxtController.java
```diff
@@ -194,31 +197,32 @@ public void move(MotorDir leftMotorDir, MotorDir rightMotorDir) {
...
-     * @param percent
+     * @param power
      */
-    public void setMotorPower(MotorKind kind, double percent) {
-        if (percent > 100.0) {
-            percent = 100.0;
-        } else if (percent < 0) {
-            percent = 0.0;
+    public void setMotorPower(MotorKind kind, int power) {
+        if (power > 100) {
```
mandaiy, line 208: I'm wondering whether this clipping is good for users of this class.
Which way do you think better, throwing exceptions or clipping values?
Anyway, at least we should note the way in doc.

app/src/nxt/java/com/pileproject/drive/execution/NxtController.java
```diff
@@ -194,31 +197,32 @@ public void move(MotorDir leftMotorDir, MotorDir rightMotorDir) {
...
                 break;
             case RightMotor:
-                mRightMotorPower = (int) ((MAX_MOTOR_POWER * percent) / 100.0);
+                mRightMotorPower = power;
                 break;
         }
     }
 
     @Override
     public void finalize() throws RuntimeException {
```
mandaiy, line 225: We must change this; read [this](http://www.informit.com/articles/article.aspx?p=1216151&seqNum=7).
Maybe it's good to do in next PR.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,206 @@
...
+        Whitebox.setInternalState(controller, "mLineSensor", lineSensor);
+
+        doReturn(20).when(lineSensor).getSensorValue();
+        assertEquals(20, controller.getLineSensorValue());
+
+        doReturn(200).when(lineSensor).getSensorValue();
+        assertEquals(200, controller.getLineSensorValue());
+
+        doReturn(-20).when(lineSensor).getSensorValue();
+        assertEquals(-20, controller.getLineSensorValue());
```
mandaiy: Same as above.

app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java
```diff
@@ -0,0 +1,206 @@
...
+        controller.moveForward();
+        verify(leftMotor).setSpeed(NxtController.MAX_MOTOR_POWER);
+        verify(rightMotor).setSpeed(NxtController.MAX_MOTOR_POWER);
+
+        controller.setMotorPower(NxtController.MotorKind.LeftMotor, 10);
+        controller.setMotorPower(NxtController.MotorKind.RightMotor, 10);
+
+        controller.moveForward();
+        verify(leftMotor).setSpeed(10);
+        verify(rightMotor).setSpeed(10);
```
mandaiy, line 199: Same as above above above.
You also need out-of-boundary case.

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
```diff
@@ -39,31 +39,30 @@
     private OutputStream mOutputStream;
     private InputStream mInputStream;
 
-    public BluetoothCommunicator(BluetoothDevice device) {
+    public BluetoothCommunicator(BluetoothDevice device) throws NullPointerException {
```
mandaiy: `@NonNull` is still informative because programmers can know that on IDE.

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,176 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+
+@RunWith(AndroidJUnit4.class)
+public class BlockProgramLogicTest {
+
+    ExecutionCondition.SelectionResult trueResult;
```
mandaiy: I think these can be final.

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        doReturn(1).when(condition).sizeOfSelectionResult(); // this means pushing a selection result
+
+        doReturn(trueBlock).when(condition).getCurrentBlock(); // return a block inside a true statement
+
+        doReturn(trueResult).when(condition).peekSelectionResult(); // return true result
+
+        // return a nearest Selection bock
+        doReturn(selectionBlock).when(condition)
+                .getNearestSelectionBlock(any(ExecutionCondition.SelectionResult.class));
```
mandaiy: If possible, we should avoid `any` because it makes test very loose.

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,176 @@
+/*
```
mandaiy, line 1: There's classic golden rule for unit testing: [Arrange-Act-Assert](http://defragdev.com/blog/?p=783).
I'd like to recommend that you should follow this.

mandaiy (changes requested):

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
tiwanari: Thanks for a very good reference :)

app/src/nxt/java/com/pileproject/drive/execution/NxtController.java
tiwanari: I see. To make matters worse, this doesn't override `Object`'s `finalize` (it is called explicitly by an Activity...). I'll fix it in other PR. Thanks!

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
tiwanari: Wow, thanks 👍

tiwanari:
Hi @myusak, 

Thank you for helpful comments :satisfied: I'll follow your advice :)

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
tiwanari: Oh, IDE says `call requires API level 19 (current min is 15)` ... :anguished:

app/src/nxt/java/com/pileproject/drive/execution/NxtController.java
tiwanari: I keep it as it is at present and added it into Javadocs.

tiwanari:
I fixed codes.

app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+
+@RunWith(AndroidJUnit4.class)
+public class BlockProgramLogicTest {
+
+    final ExecutionCondition.SelectionResult trueResult
```
mandaiy: Should be also static and private.
And is `ExecutionCondition#SelectionResult` needed to be not static class?
If not, the class should be modified.

app/build.gradle
```diff
@@ -67,24 +67,34 @@ android {
...
 
+    // android test libraries
     androidTestCompile 'com.android.support.test:runner:0.5'
     androidTestCompile 'com.android.support.test.espresso:espresso-core:2.2.2'
 
+    // mockito
+    androidTestCompile 'org.mockito:mockito-core:1.10.19'
+    androidTestCompile "${dexmakerGroupId}:dexmaker:${dexmakerVersion}"
```
mandaiy: I'm sorry and I really apologise that I changed my mind,
that it would be rather good not to use `initMocks` than to depend on non-release version, I think.
But If you really want to use that method, I won't complaint it.
In that case you should add a note why we are using this version and delete three commented-out lines.

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
mandaiy: I really hate this kind of boilerplates.
Why not consider using [Google Guava](https://github.com/google/guava)'s [precondition package](https://github.com/google/guava/wiki/PreconditionsExplained).

The cons of using this is that APKs using this library tend to be fat because the library has tons of packages.
I didn't check that the proguard can reduce the size of the APKs, but I believe it is worth considering the use
of the library in future.

Anyway I should conclude that Android API is a mess.

mandaiy (changes requested):
Well, I found one minor thing in test and one problem on the build.gradle.
Please reply about them.

app/src/main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
tiwanari: Thanks for a nice suggestion :) It may have more power than I expected and introducing it looks really valuable for this product. However, I think we should consider the cons compared to its pros - I want to know the pros more - and, thus, it should be done in a substantial PR with making the most of the library - so let's exploit it not only for this kind of small thing.

tiwanari:
Hi @myusak,

I modified codes based on your helpful comments :star:
I keep some topics discussed here as cards in [Release](https://github.com/PileProject/drive/projects/1) project such as [Guava](https://github.com/PileProject/drive/projects/1#card-551784) and [Finalizer](https://github.com/PileProject/drive/projects/1#card-551781).

tiwanari (commented):
Sorry for disturbing you. I'm not really familiar with the new GitHub review process. I replied to one of your comments (about Guava) but it doesn't appear without submitting this.

mandaiy (approved):

PR #48, opened 2016-10-19
Upgrade SquiDB from 2.0.2 to 3.1.3

tiwanari:
I upgrade SquiDB. Most of the changes are based on some class name changes.
### Note: `_id` generation and warnings

SquiDB is in a transitional period as for `_id` column which is automatically generated by default but it will be disabled in the future and now it produces `getId() is deprecated` warnings after 3.1.0 even if follows their instruction (So, please ignore these warnings).

For more detail, please see: [Primary keys · yahoo/squidb Wiki](https://goo.gl/xXK9er)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -67,6 +67,8 @@ android {
     }
 }
 
+def squidbVersion = '3.1.3'
+
 dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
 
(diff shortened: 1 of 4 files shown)


tiwanari:
I'm sorry for lacking a reference to a document I followed: 
https://github.com/yahoo/squidb/wiki/Changes-in-SquiDB-3.0

mandaiy (approved):
looks good. good job.

PR #38, opened 2016-09-25
Allow to choose the firmware of nxt

tiwanari:
This PR provides a feature for users to select the firmware of NXTs. In code level, the difference of these firmwares is just the output value of TouchSensor (it is opposite actually) in this application. 

The following window is what this PR shows to users.
screenshot_20160925-220618

PR #51, opened 2016-10-23
Hotfix/setting bugs

### LINKED ISSUE #43: Sub-screens in Setting are collapsed
Discussed on #38

tiwanari:
I found the port setting didn't work properly. So, I made this PR.

I think this will also close #43. I shortened the name of devices (i.e., sensors and motors) in order not to make screen collapsed. Please look at the following figure.

BTW, @myusak said the threshold sub-screen also collapsed on #38 but I couldn't see that. So, could you share a figure that illustrates it? :)

screenshot_20161023-140749

### DIFF
--- a/app/src/main/java/com/pileproject/drive/view/PortTextViewBase.java
+++ b/app/src/main/java/com/pileproject/drive/view/PortTextViewBase.java
@@ -31,8 +31,8 @@ import com.pileproject.drive.R;
  * A TextView which expresses a port.
  */
 public abstract class PortTextViewBase extends TextView {
-    final private String mPortName;
-    final private String mPortType;
+    private final String mPortName;
+    private final String mPortType;
     protected String mDeviceType;
 
     public PortTextViewBase(Context context, AttributeSet attrs) {
--- a/app/src/main/java/com/pileproject/drive/view/PortTextViewBase.java
+++ b/app/src/main/java/com/pileproject/drive/view/PortTextViewBase.java
@@ -31,7 +31,6 @@ import com.pileproject.drive.R;
  * A TextView which expresses a port.
  */
 public abstract class PortTextViewBase extends TextView {
-    final private boolean mIsFreePort;
     final private String mPortName;
     final private String mPortType;
     protected String mDeviceType;
@@ -42,7 +41,6 @@ public abstract class PortTextViewBase extends TextView {
         TypedArray tar = context.obtainStyledAttributes(attrs, R.styleable.PortTextViewBase);
         mPortName = tar.getString(R.styleable.PortTextViewBase_portName);
         mPortType = tar.getString(R.styleable.PortTextViewBase_portType);
-        mIsFreePort = (mPortName == null); // NOTE: a free port has no name
         tar.recycle();
 
         setOnTouchListener(new OnTouchListener() {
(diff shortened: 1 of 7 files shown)


mandaiy:
This was what I got in my emulator (Nexus_5X_API_24_x86)
I didn't check that screen resolution do matter or not, the above was in 1080x1920, 420dpi

device-2016-10-25-203231

tiwanari:
Hi @myusak,
Thanks! I checked it with different resolutions & APIs and found it may be a problem related to APIs because I got it with API 23+ and didn't see with API 22 regardless of resolutions (Phone, Tablet).
Anyway, I will dig into the problem :)

mandaiy:
Or we can solve this by adding resizing function like [this](https://github.com/PileProject/drive/blob/develop/app/src/main/java/com/pileproject/drive/setting/machine/BluetoothMachineSelectFragment.java#L185-L198)

app/src/main/res/values/strings.xml
```diff
@@ -70,7 +70,7 @@
 
     <!-- TextsOnBlock -->
     <!-- Sequence -->
-    <string name="textOnBlocks.contentDefaultSec">0</string>
+    <string name="textOnBlocks.contentDefaultSec">0.0</string>
```
mandaiy: well, this should be `0` because if the locale of the system is, for example `France`, app will crash 
since French people use a comma as a decimal point.
I apologise that I didn't comment that things.

app/src/main/java/com/pileproject/drive/view/PortTextViewBase.java
```diff
@@ -31,7 +31,6 @@
  * A TextView which expresses a port.
  */
 public abstract class PortTextViewBase extends TextView {
-    final private boolean mIsFreePort;
     final private String mPortName;
```
mandaiy: git blame maybe show my name but these two should be `private final`. Thanks.

mandaiy (changes requested):
Found one critical issue and one minor thing.

app/src/main/res/values/strings.xml
tiwanari: Oh, I see... Thanks for pointing out.

Please let me leave a note for this. Suppose the following situation; 
1. The initial value on `TextView` (`NumberTextHolder`) is `0`
2. A user long touched (the trigger of `TextView`) the view
3. The user doesn't change the value and click `OK`
4. Then the value becomes `0.000` (yeah, I should have changed it to `0.000` not `0.0`:P) in the current implementation

I changed it because I thought this is a bit surprising for users but you told me that caused a bug. I think changing the initial value in `NumberTextHolder`s' constructors is neither a good idea... hmm...

app/src/main/java/com/pileproject/drive/view/PortTextViewBase.java
tiwanari: I will fix them :)

app/src/main/res/values/strings.xml
mandaiy: If we add settings for the precisions of the text views, the default values have to be modified dynamically.
So I'd suggest that the default value is set in the constructor in accordance with the locale and the precision.

Could you please make an issue for that and let's close this.

mandaiy (approved):

app/src/main/res/values/strings.xml
tiwanari: Okay, thanks for your suggestion 👍

tiwanari:
Now, I'm trying to figure out the cause of the error on the threshold sub-screen. So, give me a little time.

tiwanari:
Oh, I also found a difference between API 22 and 23. 
Note:
These dialogs (sub-screens) have titles in API 22 and not in API 23.

tiwanari:
Hi @myusak,

I used [your suggestion](https://github.com/PileProject/drive/pull/51#issuecomment-256284113) and these collapsed screen has been fixed regardless of APIs :) But I'm not still sure why this happened and don't fix about the titles...

API 22:
API 22
API 23:
API 23

mandaiy:
Maybe theme related problems?
Can you specify the dialog theme of the app in some xml file for `NoTItle` or something?

tiwanari:
I tried changing `WindowNoTitle` flag from `false` to `true` at [here](https://github.com/PileProject/drive/blob/develop/app/src/nxt/res/values/styles.xml#L40) and emulators got an error when the app starts...
Do you have any idea why this happens and how to solve it?

tiwanari:
Hmm... It seems to be a problem related to `AppCompat` updates. 
Now my question has changed from "why don't titles appear in API 23+" to "why do titles apper in API 22" in spite of the flag `windowNoTitle`.

But I think this is not so critical at this point.

tiwanari:
I will merge this

PR #52, opened 2016-10-25
Introduce Dagger2

### LINKED ISSUE #31: Transfer the responsibility of maintaining keys to the classes that use the keys
We are discussing who should keep `keys` in #30 and considering the way of launching an activity that needs keys as arguments.

### LINKED ISSUE #45: Proposal of Introducing DI
This is a proposal issue for introducing DI container (Dependency Injection)
For diverse opinions, please feel free to leave any comment, @amiq11. 
## TL;DR
- I hate the current structure of `...ActivityBase` and `Flavor...Activity`
- I want to restructure them by splitting `...ActivityBase` into common logic and flavor-specific logic
- I want to introduce Dagger2
## Background

Currently we are going to implement [product flavors](https://developer.android.com/studio/build/build-variants.html#product-flavors).
There are two main features of product flavors: determining java files and merging xml.

(You can skip below 2 sections if you know product flavors)
### determining java files

With product flavors, you can change files (or classes) by placing corresponding directories;
suppose you define 2 flavors (`flavor1`, `flavor2`) and have `flavor1/java/com/example/Test.java`, and `flavor2/java/com/example/Test.java`,
and the code in `main/java/com/example/Main.java` refers to `Test` class,
then the `Test` class is referred with accordance of the product flavor in build-time.
### merging xml

You can define `AndroidManifest.xml` files for each product flavor and one `main`.
`main` does not belong to flavors, but it defines common resources (xml, java, images).
If you have `AndroidManifest.xml` in `main` and `flavor1`, Android Studio merges the two xml files in build-time.
Why this is useful is that you can change resources according to the flavor.
A launch activity is defined in the manifest xml file. So you can choose the launch activity
by defining the manifest xml in the flavor.
### Why we use it and How we used it

Why this feature is needed for us is because:
- we have to support multiple types of robots, each of which needs slightly different code (e.g., types of block, types of communication methods)
- but code-duplication has to be minimized

That's why we are going to use productFlavors.
We have `nxt`, `ev3`, and `pile` robots so we'll have 3 product flavors.

In our current code, we  we define `TitleActivity` for each flavor and specify launch activities in manifest xml files for each flavor.
With the xml files, the launch activity of each app can start properly.
And for programming (block-manipulation) and execution (communication with robots) activities,
we define `ProgrammingActivityBase` and `ExecutionActivityBase` respectively,
and then in each flavor these classes are inherited according to the flavor (e.g., for `nxt`, `NxtProgrammingActivity` and `NxtExecutioinActivity`).

The code of `ProgrammingActivityBase` is like below:

``` java
public abstract class ProgrammingActivityBase extends Activity {

    /* these methods are implemented in concrete classes */
    protected abstract Intent getIntentForBlockList();

    protected abstract Intent getIntentForExecution();

    /* the rest is common for all flavors */
    private void goToExecution() {
        Intent intent = getIntetnForExecution();
        ...
    }
    ...
};
```

Then in `NxtProgrammingActivity` is like below:

``` java
public class NxtProgrammingActivity extends ProgrammingActivityBase {

    @Override
    protected Intent getIntentForBlockList() {
        return new NxtBlockIntent();
    }

    @Override
    protected Intent getIntentForExecution() {
        return new NxtExecutionIntent();
    }
}
```
## Problem Statement

What I am concering is that the structure seems awkward and the base classes are too huge for abstract class.
In my opinion, the straightforward way for this situation is that defining interface for flavor-specific logic (intent-provider for programming)
and injecting flavor-specific-implementation into the interface.

So what I want to suggest is like below:

``` java
public class ProgrammingActivity extends Activity {
    private IntentProvider intentProvider;

    private void goToExecution() {
        Intent intent = intentProvider.forExecution();
        ...
    }
}

public interface IntentProvider {
    Intent forExecution();
}

public class NxtIntentProvider {
    @Override
    public Intent forExecution() {
        return ...
    }
}
```

By splitting code like above, the code gets
- better organized
- easier to be mocked
- easier for new-comers to understand what is needed to be implemented for other flavors.

And this proposal also solves #31.
### Coping with dependency by DI container

If you feel my proposal sounds good, you might wonder how the interfaces are `injected`.
That kind of problems are called Dependency Injection.
There are a number of pages describing this problem, so please google it.

For Android projects, there is a cool DI container called [Dagger2](http://google.github.io/dagger/).
So I want to use this for our android project.

I tested the feasibility of my proposal in [here](https://github.com/myusak/DaggerTest)

By using Dagger2, the code would be like below:

``` java

public class ProgrammingActivity extends Activity {

    @Inject
    IntentProvider intentProvider;

    /* same as above */
}
```

How cool, isn't it?
Thank you for reading this too long issue.

mandaiy:
This PR closes #31 and #45.

Following changes are made:
- Renamed Activities' name by removing suffix "Base"
- Changed intent-creation strategy. An intent for an Activity is created by the Activity's static method
- Added DI code for BlockProvider and MachineProvider

In addition to these changes, minor changes were made
- Moved Block category constants to `@BlockCategory` annotation interface
- Changed the way of invoking MachineBase.connect (Thread to RxJava) in ExecutionActivity
- Refactored some code


tiwanari:
Hi @myusak,

I haven't looked into your codes yet but I guess this PR also resolves #31 thanks to the change on how to make intents;

> Changed intent-creation strategy. An intent for an Activity is created by the Activity's static method

Is my guess right?

Will you add more changes? Please let us know when you finished fixing :)

app/src/main/java/com/pileproject/drive/programming/visual/block/BlockCategory.java
```diff
@@ -0,0 +1,42 @@
...
+
+import static com.pileproject.drive.programming.visual.block.BlockCategory.REPETITION;
+import static com.pileproject.drive.programming.visual.block.BlockCategory.SELECTION;
+import static com.pileproject.drive.programming.visual.block.BlockCategory.SEQUENCE;
+
+/**
+ * Annotation that represents a category of the blocks in this App.
+ * <ul>
+ *     <li><code>SEQUENCE</code>: Sequence block</li>
+ *     <li><code>REPETITION</code>: REPETITION block</li>
```
tiwanari: nit: the second `REPETITION` should be `Repetition`

mandaiy:
your understanding is correct and i won't add commits anymore unless you want

tiwanari:
Thanks for your confirmation :)
I will check them soon 👍

app/src/main/java/com/pileproject/drive/programming/visual/activity/BlockListActivity.java
```diff
@@ -21,126 +21,194 @@
...
-        public BlockClassHolder(Class<? extends BlockBase> clazz) {
+        BlockClassHolder(Class<? extends BlockBase> clazz) {
             mClass = clazz;
         }
 
-        public String getDescription() {
+        String getDescription() {
             String className = mClass.getSimpleName();
-            StringBuffer resourceName = new StringBuffer(className);
+            StringBuilder resourceName = new StringBuilder(className);
```
tiwanari, line 212: This is good. So, why not apply this change to `getIconResourceId`?

app/src/main/java/com/pileproject/drive/util/bluetooth/RxBluetoothConnector.java
```diff
@@ -77,11 +81,47 @@ public void call(Subscriber<? super BluetoothSocket> subscriber) {
 
                     subscriber.onNext(socket);
                     subscriber.onCompleted();
-                } catch (IOException e) {
-                    subscriber.onError(e);
+                } catch (IOException firstCause) {
+
+                    try {
+                        BluetoothSocket socket = connectByRfcommSocket(bluetoothDevice);
+                        subscriber.onNext(socket);
```
tiwanari, line 92: Isn't `subscriber.onCompleted()` necessary after this line?

app/src/main/java/com/pileproject/drive/util/bluetooth/RxBluetoothConnector.java
```diff
@@ -77,11 +81,47 @@ public void call(Subscriber<? super BluetoothSocket> subscriber) {
...
+                    } catch (RfcommSocketInvocationException e) {
+                        subscriber.onError(firstCause);
```
tiwanari: You may as well mention which method (first or second) failed to connect. Users can't see which line (this or L. 92) passes the exception to the subscriber.

app/src/main/java/com/pileproject/drive/execution/ExecutionActivity.java
```diff
@@ -135,14 +152,25 @@ public void onClick(View v) {
...
 
-        if (hasBluetoothFunction()) {
-            if (!mBtAdapter.isEnabled()) {
-                startActivityForResult(new Intent(BluetoothAdapter.ACTION_REQUEST_ENABLE),
-                                       REQUEST_ENABLE_BT);
-            } else {
-                connectToDevice();
-            }
+        // TODO: remove bluetooth-dependent code for WifiCommunicator
+        if (!DeployUtils.isOnEmulator() && !BluetoothUtil.hasBluetoothFunction()) {
```
tiwanari, line 159: How about making an issue for this as a note so that we can remember we have a plan to prepare `WifiCommunicator`?

tiwanari (approved):
There are some nits but almost all look pretty good to me!

app/src/main/java/com/pileproject/drive/execution/ExecutionActivity.java
mandaiy: gotcha.

app/src/main/java/com/pileproject/drive/programming/visual/activity/BlockListActivity.java
mandaiy: Well, I missed it. Fixed.

app/src/main/java/com/pileproject/drive/programming/visual/block/BlockCategory.java
mandaiy: Fixed.

app/src/main/java/com/pileproject/drive/util/bluetooth/RxBluetoothConnector.java
mandaiy: You're right haha.

By the way this method is not used at this point; I didn't intend to include this file 
because I couldn't modify BluetoothCommunicator#open so as to use this method.
Do you think this hunk is needed for this PR?

app/src/main/java/com/pileproject/drive/util/bluetooth/RxBluetoothConnector.java
mandaiy: Added a message for the Throwables.

app/src/main/java/com/pileproject/drive/util/bluetooth/RxBluetoothConnector.java
tiwanari: Oh, I missed the point. So, what is the difficulty to use this method in `BluetoothCommunicator#open`? I think a note for that will help us to consider whether this method is necessary or not in the future :)

tiwanari: Oh, I missed the point. So, what is the difficulty to use this method in(?) BluetoothCommunicator#open? I think a note for that will help us to consider whether this method is necessary or not in the future :)

mandaiy: In `RxBluetoothCommunicator#connect`, if the connection fails, it calls `subscriber#onError` method with the cause (`Throwable`).
Such a `onError` runs on a different scope from `BluetoothCommunicator#open`.
While `BluetoothCommunicator#open` is supposed to throw `IOException` if the connection fails, `Subscriber#onError` _cannot_ pass the cause of error to the `BluetoothCommunicator#open`.
In short, `BluetoothCommunicator#open` cannot throw exceptions if failed.

I think we do not have to stick to Rx way for `BluetoothCommunicator`, 
we do not have to modify the current implementation.

tiwanari: I see and agree to keep the current implementation. I think keeping its simplicity as an example is also helpful for other developers :)

tiwanari:
Hi @amiq11,

Though I approved this PR, please have a look and feel free to make comments :) If you don't have enough time, please let us know.

mandaiy:
@amiq11 ping.
I gonna merge this until the end of tomorrow unless you complain. Thanks.

PR #55, opened 2016-10-30
Add RxObservableProgram for program execution

### LINKED ISSUE #33: Do not use raw Thread
We use a raw Thread to execute programs (See [ExecutionThread](https://github.com/PileProject/drive/blob/develop/app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java)). We plan to replace it with another modern way (TBD).

mandaiy:
This commit removes ExecutionThread class.
Instead of the class, RxObservableProgram is added,
which is Rx version of the Thread class.
This commit also renames the name of `MachineController#finalize' to`MachineController#close'

Changes of ExecutionActivity are also made in this commit.

Closes #33


mandaiy:
Note that despite the implementation is exactly same as the ExecutionThread, 
the execution will be stopped unexpectedly if you
1. stop (pause in this commit-term) the program and then
2. restart the program
The cause was accessing out of index in ExecutionCondition.

mandaiy:
> Note that despite the implementation is exactly same as the ExecutionThread, 
> the execution will be stopped unexpectedly if you
> 1. stop (pause in this commit-term) the program and then
> 2. restart the program
> The cause was accessing out of index in ExecutionCondition.

Problem solved. The cause was that `decrementProgramCount` was called multiple times in [here](https://github.com/PileProject/drive/blob/develop/app/src/main/java/com/pileproject/drive/execution/ExecutionThread.java#L74)

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -0,0 +1,190 @@
...
+ * Do not run this on your UI thread.
+ *
+ * <code>
+ *     RxObservableProgram program = new RxObservableProgram(...);
+ *     Observable.create(program)
+ *         .subscribeOn(Schedulers.newThread())
+ *         .subscribe(...);
+ * </code>
+ */
+public class RxObservableProgram implements Observable.OnSubscribe<Bundle> {
```
tiwanari, line 62: We use Hungarian notation for members. So please add prefixes of members.

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    private void mainLoop(Subscriber<? super Bundle> subscriber) throws RuntimeException {
+
+        boolean haltAlreadyCalled = false;
+
+        while (!condition.hasProgramFinished()) {
+
+            if (terminateRequest) {
+                subscriber.onNext(makeMessageBundle(END));
```
tiwanari: This `END` and `PAUSE` are just ignored in `ExecutionActivity`, right? Do you have any plan to use them?

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -0,0 +1,190 @@
...
+                break;
+            }
+
+            if (pauseRequest) {
+                if (!haltAlreadyCalled) {
+                    controller.halt();
+                    condition.decrementProgramCount();
+
+                    subscriber.onNext(makeMessageBundle(PAUSE));
```
tiwanari: Same above.

tiwanari (changes requested):
Totally, looks pretty good and clean :)
There is a naming style problem for consistency and a question about optional `onNext`s

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
mandaiy: I feel like this is a library-ish class so the class has to be capable of emitting any kind of events the class is responsible for.

tiwanari: Okay, it's reasonable but I think `END` is somewhat confusing because the meaning of it looks the same as `subscriber.onCompleted()`. I propose some options for this;
1. Remove `END` from this class (`onCompleted()` does the same thing)
2. Rename `END` to `TERMINATE_BY_USER` or something (`onCompleted()` notifies the normal ending)
3. Keep using this `END` but remove it from here and insert it after this while loop so that this `END` express all the end cases

tiwanari:
NOTE: I replied a comment about `END`

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
mandaiy: Okay I overlooked something, you definitely true. Option 2 seems good to me.

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -0,0 +1,214 @@
...
+ * This class is one-off.
+ * Create this instance whenever you want to execute and DO NOT REUSE the instance.
+ * <p/>
+ * The stream emits {@link Bundle} objects for the subscriber.
+ * The bundles contain message type whose key is {@link RxObservableProgram#KEY_MESSAGE_TYPE},
+ * and optionally one message argument whose key is {@link RxObservableProgram#KEY_MESSAGE_ARG}.
+ * The message types are:
+ * <ul>
+ *     <li>{@link RxObservableProgram#MESSAGE_STARTED}: emitted at the beginning of the execution</li>
+ *     <li>{@link RxObservableProgram#MESSAGE_TERMINATED}: emitted when {@link RxObservableProgram#requestRestart()} is called</li>
```
tiwanari: nit: this may be `RxObservableProgram#requestTerminate()`

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -0,0 +1,214 @@
...
+ *     <li>{@link RxObservableProgram#MESSAGE_PAUSED}: emitted when {@link RxObservableProgram#requestPause()} is called</li>
+ *     <li>{@link RxObservableProgram#MESSAGE_BLOCK_EXECUTED}: emitted when a block is executed.
+ *         this message contains an argument which is the index of the block and can be accessed
+ *         with {@link RxObservableProgram#KEY_MESSAGE_ARG} </li>
+ * </ul>
+ * When the execution of the program that this class holds is ended, {@link rx.Subscriber#onCompleted} is called,
+ * no matter the type of ending (terminated by a user/ended normally).
+ * <p/>
```
tiwanari: nit: Javadoc doesn't allow us to use self-closing elements from Java8 ( [reference](http://stackoverflow.com/questions/26049329/javadoc-in-jdk-8-invalid-self-closing-element-not-allowed) )

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -0,0 +1,214 @@
...
+ * Create this instance whenever you want to execute and DO NOT REUSE the instance.
+ * <p/>
+ * The stream emits {@link Bundle} objects for the subscriber.
+ * The bundles contain message type whose key is {@link RxObservableProgram#KEY_MESSAGE_TYPE},
+ * and optionally one message argument whose key is {@link RxObservableProgram#KEY_MESSAGE_ARG}.
+ * The message types are:
+ * <ul>
+ *     <li>{@link RxObservableProgram#MESSAGE_STARTED}: emitted at the beginning of the execution</li>
+ *     <li>{@link RxObservableProgram#MESSAGE_TERMINATED}: emitted when {@link RxObservableProgram#requestRestart()} is called</li>
+ *     <li>{@link RxObservableProgram#MESSAGE_PAUSED}: emitted when {@link RxObservableProgram#requestPause()} is called</li>
```
tiwanari, line 42: I feel it is helpful to add an explanation for `how we know an execution restarted` with this `PAUSED` message. We can know it by receiving `BLOCK_EXECUTED`.

tiwanari (approved):
I made some comments for small things. Please fix them if they look good to you. Anyway, I approve this PR.
Overall, they remove ugly logics and now we have more modern codes :) Great job! 👍

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
mandaiy: Fixed

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
mandaiy: Added MESSAGE_RESTARTED

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
mandaiy: Fixed

mandaiy:
@tiwanari I added some changes so if you have time please take a look on it.

tiwanari:
FYI, you may dismiss a review by following [this article](https://help.github.com/articles/dismissing-a-pull-request-review/) to request another review when you modify codes significantly. This is a somewhat dangerous feature though.

tiwanari (commented):
I think `commitAllowingStateLoss` is not bad at least at this point.
Thanks for fixing a bug on communication. I will check it with a real robot and report some results by tomorrow's morning.

tiwanari:
Hi @myusak, 

I checked it and simple programs can be executed properly :)

But I have a trouble;

1. When a program execution finishes, I get a message something like `execution is over`
2. Before pushing `OK`, go home screen with the home button
3. Go back to our app
  - At this time, I am in the programming screen
4. Try to execute other problem
5. Then I always get `Connection error` and can't run any programs

tiwanari:
The AVD monitor says

```
11-03 14:05:15.467: D/BluetoothCommunicator(15368): Failed to connect with an orthodox method
11-03 14:05:15.467: D/BluetoothCommunicator(15368): Failed to connect with a redundancy method
```

So, maybe `connect()` always fails in such case. It doesn't do `close()` or something like that?

tiwanari:
Oh, I'm very sorry. The problem I described in previous two comments (UPDATE: I deleted them) was caused by a problem of mine. I checked they work correctly for the cases with the latest commit 👍 

By the way, I found our app crashes when I enable Bluetooth function after pushing `Execute` button. If the Bluetooth function is off, the app asks me to turn on the Bluetooth and I push `OK` then I get the message "NxtDrive crashed unexpectedly."

I dug into this and found this is caused by `ProgressFragment`. Please see the following log;

```
11-04 00:30:50.975: E/AndroidRuntime(23044): FATAL EXCEPTION: main
11-04 00:30:50.975: E/AndroidRuntime(23044): java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=1, result=-1, data=null} to activity {com.pileproject.drive.nxt/com.pileproject.drive.execution.ExecutionActivity}: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.ActivityThread.deliverResults(ActivityThread.java:3071)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.ActivityThread.handleSendResult(ActivityThread.java:3114)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.ActivityThread.access$1100(ActivityThread.java:127)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1211)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.os.Handler.dispatchMessage(Handler.java:99)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.os.Looper.loop(Looper.java:137)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.ActivityThread.main(ActivityThread.java:4558)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at java.lang.reflect.Method.invokeNative(Native Method)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at java.lang.reflect.Method.invoke(Method.java:511)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:784)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:551)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at dalvik.system.NativeStart.main(Native Method)
11-04 00:30:50.975: E/AndroidRuntime(23044): Caused by: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:1493)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.support.v4.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:1511)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.support.v4.app.BackStackRecord.commitInternal(BackStackRecord.java:638)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.support.v4.app.BackStackRecord.commit(BackStackRecord.java:617)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.support.v4.app.DialogFragment.show(DialogFragment.java:139)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at com.pileproject.drive.util.fragment.ProgressDialogFragment.showDialog(ProgressDialogFragment.java:53)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at com.pileproject.drive.execution.ExecutionActivity.connectToMachine(ExecutionActivity.java:234)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at com.pileproject.drive.execution.ExecutionActivity.onActivityResult(ExecutionActivity.java:282)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.Activity.dispatchActivityResult(Activity.java:4839)
11-04 00:30:50.975: E/AndroidRuntime(23044):    at android.app.ActivityThread.deliverResults(ActivityThread.java:3067)
11-04 00:30:50.975: E/AndroidRuntime(23044):    ... 11 more
```

Could you fix it in this PR?

mandaiy:
> The problem I described in previous two comments was caused by a problem of mine.

What was your problem? just curious.

> I found our app crashes when I enable Bluetooth function after pushing Execute button. If the Bluetooth function is off, the app asks me to turn on the Bluetooth and I push OK then I get the message "NxtDrive crashed unexpectedly."

Might be same cause as the AlertFragmentDialog's one.
Basically the ProgressFragments do not have to store their states so the bug could be fixed by the same approach.

tiwanari:
> What was your problem? just curious.

I had a trouble with installing this app into a device via micro-USB (I figured out that it was the problem of a micro-USB of mine), so I made an apk and installed it by using Dropbox. Then, when I tried to execute a program with the installed app, I caught that exception. But it doesn't occur when I install the app via micro-USB. 

I don't know why but I guess this will not happen in the standard way (e.g., downloading it from PlayStore). I'm really sorry for that.

tiwanari:
Here is a brief patch (diff.patch) I used to solve the bug;

``` patch
diff --git app/src/main/java/com/pileproject/drive/util/fragment/ProgressDialogFragment.java app/src/main/java/com/pileproject/drive/util/fragment/ProgressDialogFragment.java
index 9446d38..03d8d8e 100644
--- app/src/main/java/com/pileproject/drive/util/fragment/ProgressDialogFragment.java
+++ app/src/main/java/com/pileproject/drive/util/fragment/ProgressDialogFragment.java
@@ -23,6 +23,7 @@ import android.os.Bundle;
 import android.support.annotation.StringRes;
 import android.support.v4.app.DialogFragment;
 import android.support.v4.app.FragmentManager;
+import android.support.v4.app.FragmentTransaction;
 import android.text.TextUtils;

 /**
@@ -50,7 +51,10 @@ public class ProgressDialogFragment extends DialogFragment {
      */
     public static void showDialog(FragmentManager manager, String title, String message, String tag) {
         ProgressDialogFragment f = newInstance(title, message);
-        f.show(manager, tag);
+
+        FragmentTransaction transaction = manager.beginTransaction();
+        transaction.add(f, tag);
+        transaction.commitAllowingStateLoss();
     }

     /**
```

Please use this with the following command in your `drive` repository;

```
patch -p0 < diff.patch
```

mandaiy:
Hey @tiwanari Sorry for my late reply.
Applied your patch and if Travis approve the patch, I'll merge this.
Thanks.

tiwanari:
It's all right! :)
This PR looks so nice! Thanks, too! 👍

Issue #40, opened 2016-09-26
Add Copyright Notice


tiwanari:
The drive application does not have a copyright notice. 

I will gather notices from external libraries and show them in SettingActivity.

tiwanari:
Here is a list  of external libraries.
- [ ] com.pileproject:drivecommand:2.1.0
- [ ] com.android.support:support-v4:23.2.1
- [ ] com.android.support:appcompat-v7:23.4.0
- [ ] com.android.support:design:23.2.1
- [ ] com.android.support:preference-v7:23.0.1
- [ ] com.yahoo.squidb:squidb:2.0.2
- [ ] com.yahoo.squidb:squidb-annotations:2.0.2
- [ ] com.rejasupotaro:kvs-schema:5.0.0
- [ ] co.trikita:log:1.1.5
- [ ] io.reactivex:rxandroid:1.2.1
- [ ] io.reactivex:rxjava:1.1.6

tiwanari:
I will leave references for LICENSEs.
- [drivecommand/LICENSE](https://goo.gl/PO7bPH)
- [squidb/LICENSE](https://goo.gl/cUzTkc)
- [kvs-schema/LICENSE.txt](https://goo.gl/Vaj6tG)
- [log/LICENSE](https://goo.gl/ksQp4a)
- [RxAndroid/LICENSE](https://goo.gl/xOk9cI)
- [RxJava/LICENSE](https://goo.gl/R5g3he)

Unknown:
- [android/platform_frameworks_support](https://goo.gl/WRvwXk)

tiwanari:
[WIP] I listed LICENSEs.

tiwanari:
I searched for a good way to show them and found some articles so I leave them here.

### Collect manually and show them as Dialog (or WebView)
- (Japanese) [DialogFragmentで作るライセンス情報 - Qiita](https://goo.gl/Ub4vDq)
- ["Open source licenses" screen](https://goo.gl/NC4YMr)

### Collect automatically
- [hierynomus/license-gradle-plugin: Manage your license(s)](https://goo.gl/nXLLvi)
  - (Japanese) to know how to use the above plugin, see: [build.gradleのdependenciesに追加されたライブラリのライセンス情報を出力する - woshidan's blog](https://goo.gl/PJKQsk)

UPDATE:
I think we can use both of them. 
1. collect copyrights automatically with the bottom way
2. show them with the above way

tiwanari:
[...]

PR #57, opened 2016-11-08
Add Copyright Notice

tiwanari:
This PR solves #40.

This PR introduces ;

- the copyright notice of libraries used in this app
- a dialog fragment in preferences to show a WebView that loads the copyright notice as a html (`licenses.html`)
- a script to update the `licenses.html` easily


tiwanari:
We can update the `license.html` by using `scripts/update_licenses.sh`. 
At this time, the app shows the following dialog when you click the menu `Copyright Notice`.
screenshot_1478586986

I know this is very ugly and it contains some meaningless rows (no licensed). I will create a mechanism to format the file in a good style with the script. I'm using a plugin to collect licenses I referred in #40. 

NOTE: The plugin fails to collect licenses of some libraries such as https://github.com/rejasupotaro/kvs-schema. I'm not sure why (maybe, the LICENSE file should not have any extension?).

tiwanari:
And I found this PR causes an error in Travis because of some tasks introduced by the plugin. 
Anyway, this PR is WIP. 

Please give this some comments. What do you think about this way to update and show copyrights?

mandaiy:
I don't know but I don't think the copyright notices have to have pretty appearance.
(you'll find, for instance, inbox has very ~~ugly~~ simple copyright notices)

Also I can't understand why you introduced the simple shell script. Does it have any advantages against grande scripts? or you couldn't do something in gradle?

tiwanari:
Oh, I think it's better to use gradle :) I'm not so familiar with making gradle scripts but I'll try. Thanks for your comment.

I'm thinking of making a placeholder of copyright such as https://gist.github.com/cyrilmottier/4124374, and generating a html by inserting the libraries' information collected by a plugin. 
My first idea to do that was making a simple Ruby script because it's easy and there is no other reason at present.

tiwanari:
I changed my mind to use [another plugin](https://github.com/cookpad/license-tools-plugin) to make `licenses.html` and the plugin generates the HTML semi-automatically. 

## How to use it for **others'** licenses
Firstly, run

```
# ./gradlew checkLicenses
```

and you'll see some lists of licenses we use if they are lack in `app/licenses.yml`. The list is somewhat incomplete because some elements are not filled. So, you should gather the lacked information from their pages and fill it. Then, run

```
# ./gradlew generateLicensePage
```

If there is no error, you can get `app/src/main/assets/licenses.html`

tiwanari:
While I decided to use the another plugin for others' libraries, I keep [the original one](https://github.com/hierynomus/license-gradle-plugin) for **our** license. Because it can maintain all the license headers of source files with a command.

## How to use it for **our** license
Keep `LICENSE` file in the root directory that should be placed in the headers of source files.

Run 
```
# ./gradlew licenseFormat
```
and this will format all the files.

If you want to know all the files follow the `LICENSE` file, run

```
# ./gradlew license
```
and this will be executed automatically while building our app.


### Note
I'm not sure why it happens but `*Test.java` violates the license...

See [the build error](https://travis-ci.org/PileProject/drive/builds/178509265);
```
> License violations were found: /home/travis/build/PileProject/drive/app/src/androidTest/java/com/pileproject/drive/execution/NxtControllerTest.java,/home/travis/build/PileProject/drive/app/src/androidTest/java/com/pileproject/drive/execution/BlockProgramLogicTest.java,/home/travis/build/PileProject/drive/app/src/androidTest/java/com/pileproject/drive/programming/visual/block/NumberTextViewDelegateTest.java,/home/travis/build/PileProject/drive/app/src/androidTest/java/com/pileproject/drive/util/development/UnitTest.java,/home/travis/build/PileProject/drive/app/src/androidTest/java/com/pileproject/drive/comm/BluetoothCommunicatorTest.java,/home/travis/build/PileProject/drive/app/src/androidTest/java/com/pileproject/drive/ApplicationTest.java}
```
To avoid this, I set `exclude` for `*Test.java`.
```
license {
    include "**/*.java"
    exclude "**/*Test.java"
}
```

tiwanari:
And now, you can see the others' license like this;

screenshot_1479966966

(It is better than inbox's one, isn't it? :P )

tiwanari:
Since the license checks of new plugins will be executed automatically, I had to fix the license headers of files and they may be annoying when you review this PR. I'm sorry for that.

tiwanari:
Ah... one more thing. I had to fix `.travis.yml` because it became to always fail. I specified a supported android sdk version in it.

tiwanari:
Hey @myusak,

Please review this PR. Thank you in advance.

mandaiy (approved):

PR #59, opened 2016-11-29
Standardize controllers

tiwanari:
This PR standardizes the `MachineController` to make it has the union of its children's methods.

## Background
We talked over the design of controllers again and again offline, and we had two options;

- Make controllers for each machine (Nxt, Ev3, Pile, etc.) so that they have their specific methods (e.g., Pile Robot doesn't have `ColorSensor`, thus it doesn't have `getColorSensorValue` or something like that) and keep the `MachineController` as it is.
- Make the common `MachineController` that has the union of methods of these controllers.

They actually have pros & cons and our opinions conflicted. But finally, at present, we decided to choose the latter.

#### Pros
- Straightforward way and thus readable. This is very important for developers.
- They can be easily tested.

#### Cons
- There will be many duplications of codes because these controllers use [our library](https://github.com/PileProject/drivecommand)'s `MachineBase` that provides the same interface to control machines (so, we will have very similar codes in blocks like [ForwardSecBlock](https://github.com/PileProject/drive/blob/develop/app/src/nxt/java/com/pileproject/drive/programming/visual/block/sequence/ForwardSecBlock.java) in each build variant) but we currently can't make the most of it because of the design.
- It doesn't follow the design of [our library](https://github.com/PileProject/drivecommand). **But we can fix that of our library easily.**

#### Pros
- It follows the design of [our library](https://github.com/PileProject/drivecommand).
- It commonizes the blocks through `MachineController` interface. Thus, it reduces the duplicated codes not only `*.java` but also `*.xml`. 
- Easy to maintain because the codes are unified. 

#### Cons
- It is difficult to test each variant.
- Not straightforward way. 
- Users can't know what kinds of methods each controller has, and make it worse, calling wrong methods lead to `RuntimeError` not `CompileError`.
- Poor extensibility. We should add/remove a method to/from `MachineController` itself if new machines with new methods. So, maybe we should make an another layer for other machines something like `MachineController` -> `CarController`, `DroneController` (i.e., not use the `MachineController` as the current `NxtController`'s parent).

## Why we chose the latter
This was a very difficult problem for us but we took the pros of the latter. Maintaining duplicated codes `*.java` and `*.xml` is very annoying and we can easily change our mind when a new type of machine comes if we keep the unified codes clean.


app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
...
+import com.pileproject.drivecommand.machine.device.output.Buzzer;
+import com.pileproject.drivecommand.machine.device.output.Led;
+import com.pileproject.drivecommand.machine.device.output.Motor;
+import com.pileproject.drivecommand.machine.device.port.InputPort;
+import com.pileproject.drivecommand.machine.device.port.OutputPort;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class CarControllerBase implements MachineController {
```
tiwanari: This will be the base class for `NxtController`, `Ev3Controller` and `PileController`

.idea/gradle.xml
```diff
@@ -13,12 +13,7 @@
             <option value="$PROJECT_DIR$/app" />
           </set>
         </option>
-        <option name="myModules">
-          <set>
-            <option value="$PROJECT_DIR$" />
-            <option value="$PROJECT_DIR$/app" />
-          </set>
-        </option>
+        <option name="resolveModulePerSourceSet" value="false" />
```
tiwanari, line 16: Sorry, I don't know what it is.

app/src/main/java/com/pileproject/drive/execution/MachineProvider.java
```diff
@@ -38,7 +38,7 @@
      * Returns a concrete machine controller for the given machine.
      *
      * @param machineBase instance of {@link MachineBase} which will be manipulated by the returned {@link MachineController}
-     * @return {@link MachineController} instance
+     * @return {@link CarControllerBase} instance
      */
-    MachineController getMachineController(MachineBase machineBase);
+    CarControllerBase getMachineController(MachineBase machineBase);
```
tiwanari: Hmm, this may be something weird considering the name of this class. WDYT?

app/src/main/java/com/pileproject/drive/execution/RxObservableProgram.java
```diff
@@ -69,7 +69,7 @@
     public static final int MESSAGE_RESTARTED = 4;
     public static final int MESSAGE_BLOCK_EXECUTED = 5;
 
-    private final MachineController mMachineController;
+    private final CarControllerBase mMachineController;
```
tiwanari: Here, this forces us to use `CarControllerBase`. It may limit the extensibility.

app/src/main/java/com/pileproject/drive/preferences/BlockPreferencesSchema.java
```diff
@@ -20,15 +20,13 @@
 
 /**
  * Schema for "block_preferences" table
- *
- * @author <a href="mailto:tatsuyaw0c@gmail.com">Tatsuya Iwanari</a>
- * @version 1.0 21-April-2016
  */
 @Table(name = "block_preferences")
 public class BlockPreferencesSchema {
```
tiwanari, line 25: This class had to be moved into `main` because the thresholds are used in the `selection` blocks. But what is the `default` values for machines. They have different defaults.

app/src/main/java/com/pileproject/drive/preferences/BlockPreferencesSchema.java
```diff
@@ -20,15 +20,13 @@
...
     public class THRESHOLD {
         public static final int SOUND_DEFAULT = 70;
         public static final int LINE_DEFAULT = 50;
+        public static final int _DEFAULT = 50;
```
tiwanari: I should add the thresholds of all the sensors.

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfMachineIsOutOfLineBlock.java
```diff
@@ -18,32 +18,30 @@
...
+        super(context, R.layout.block_if_machine_is_out_of_line);
 
         mThreshold = BlockPreferences.get(context).getLineSensorThreshold();
     }
 
     @Override
     protected boolean evaluateCondition(MachineController controller) {
-        // comment is weird.
-        // getLightPercent returns tenfold value
-        return ((NxtController) controller).getLineSensorValue() > mThreshold * 10;
```
tiwanari, line 47: TODO: 
Nxt should return the normalized values not 1/10 results.

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfThereWasALargeSoundBlock.java
```diff
@@ -42,8 +42,6 @@ public IfThereWasALargeSoundBlock(Context context) {
 
     @Override
     protected boolean evaluateCondition(MachineController controller) {
-        // need multiply 10 because getdB returns 10 times value
-        // the comment is messed up
-        return ((NxtController) controller).getSoundSensorValue() > mThreshold * 10;
```
tiwanari, line 47: TODO:
Nxt should return a normalized value not 10 times results.

tiwanari (commented):
I made some changes that show briefly what we want. Do they look good? Please feel free to leave comments.

app/src/main/java/com/pileproject/drive/preferences/BlockPreferencesSchema.java
tiwanari: Ah... How about preparing some methods for getting the thresholds like `getLineSensorThreshold()` as empty (?) methods in `ControllerBase`? By overriding them in their children (i.e., `NxtController`), we can keep this class in each build variants and don't need to prepare all the threshold.

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfMachineIsOutOfLineBlock.java
tiwanari: We should add some changes in `drivecommand`

tiwanari: Oh, they were fixed in [here](https://github.com/PileProject/drivecommand/blob/develop/src/main/java/com/pileproject/drivecommand/model/nxt/NxtProtocol.java#L93). I missed it.

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfThereWasALargeSoundBlock.java
tiwanari: fixed in [here](https://github.com/PileProject/drivecommand/blob/develop/src/main/java/com/pileproject/drivecommand/model/nxt/NxtProtocol.java#L111)

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfMachineIsOutOfLineBlock.java
tiwanari: Oh, they were fixed in [here](https://github.com/PileProject/drivecommand/blob/develop/src/main/java/com/pileproject/drivecommand/model/nxt/NxtProtocol.java#L93). I missed it.

makotoshimazu:
I'm taking a look. just a moment...

app/src/main/java/com/pileproject/drive/programming/visual/activity/BlockListActivity.java
```diff
@@ -57,7 +57,11 @@
 
     public static final String KEY_BLOCK_NAME = "block-list-activity-key-block-name";
 
-    private static final int[] CATEGORY_STRINGS = {R.string.sequence, R.string.repetition, R.string.selection};
+    private static final int[] CATEGORY_STRINGS = {
```
makotoshimazu, line 60: Basically such formatting fixes especially in non-related files should be a separated issue to avoid the patch gets larger.
Could you move these updates to another cleanup PR?

tiwanari:
@amiq11 Thanks!

This is a wip PR. So please look over it and give any comment:)

tiwanari:
Notes for me;

I should...

- [x] deal with the thresholds in `BlockPreferences`
- [x] deal with `MotorProperty` and `SensorProperty` in `CarControllerBase` and `NxtCarController`
- [x] make a package for car-formed robots.

app/src/main/java/com/pileproject/drive/preferences/BlockPreferencesSchema.java
makotoshimazu: I don't understand the whole of codes, but each controller has to have methods like `bool isLineWhite()` if you embed the thresholds in the controller classes. If not, it's better to keep thresholds separated from the controllers because currently `CarController` is defined as it handles and returns values instead of boolean.

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfMachineIsOutOfLineBlock.java
```diff
@@ -18,32 +18,30 @@
...
+        return ((CarControllerBase) controller).getLineSensorValue() > mThreshold;
```
makotoshimazu: Will the downcast be removed?

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * A setter of motor power.
+     *
+     * @param kind see {@link MotorKind}
+     * @param power [0, 100]
+     *              If an out of bound value is passed, then the value will be clipped
+     *              (e.g., -10 to 0, 200 to 100).
+     */
+    public void setMotorPower(MotorKind kind, int power) {
```
makotoshimazu: I prefer `percent` to `power` because it's clearer to understand the range of the value.

makotoshimazu (commented):
I added first-round comments. I know you are still working on, so feel free to skip what I mentioned when it's not what you intended. Basically this patch looks good, but there are still several downcasts. Are you planning to remove them in another PR?

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
tiwanari: Sounds good. Thanks :)

app/src/main/java/com/pileproject/drive/programming/visual/activity/BlockListActivity.java
tiwanari: This is not just a formatting fix. I removed these `R.string.foo`s because they are duplicated and can be replaced with `R.string.programming_foo`s. Because their names are longer than the previous ones, the line became too long to keep it in one line.

app/src/main/java/com/pileproject/drive/programming/visual/block/selection/IfMachineIsOutOfLineBlock.java
tiwanari: No, it won't. Actually, this is a thing that I want to confirm. 

I prepared `CarControllerBase` for car-formed robots and intended to use these blocks that have downcasting only for these robots. I assume that there is a possibility to control other formed robots and the layer of such controllers for specific form robots is necessary. WDYT? Is it too much to do in this PR?

tiwanari: And, if I may stick to this design, do I have to move them a specific package like `com.pileproject.drive.programming.visual.block.car`?

app/src/main/java/com/pileproject/drive/preferences/BlockPreferencesSchema.java
tiwanari: Oh, I should have added more detailed explanations, sorry :bow:

There are several things I should tell you.

1. We're using `SharedPreferences` (see: [基本的なSharedPreferencesの使い方 - Qiita]( https://goo.gl/K9jz6T )) to store `Preference`.

2. We're using `kvs-schema` plugin (see: [中規模以上のアプリでSharedPreferencesをエレガントに扱う - Qiita]( https://goo.gl/t0V0WF )) to use `SharedPreferences` more elegantly.

3. `HogeSchema` is used to make an auto-generated class whose name is `Hoge` (in this case, `BlockPreferences`) with the plugin.  

4. `kvs-schema` can set the default values in `Schema` like this class. But if we keep it in `main` build variant (thus, this is used over all variants), the schema forces all apps to use the same default values for sensors in this case. But I think they may be different from robot to robot (I'm not so sure about this).

5. I want to use `BlockPreferencesSchema` in each variant independently so that they can have different defaults and this also allows us to remove useless default values for not implemented sensors (e.g., `ColorSensor` for PILE robot).

This is the background and I'm wondering what the best way is :confused:

tiwanari (commented):
@amiq11 
Thank you for your comments 👍 

I replied and some of them are what I wanted to discuss!

tiwanari:
NOTE: I checked that we can test each child controller (e.g., [NxtCarControllerTest](https://github.com/tiwanari/drive/blob/e53abf239ee03995ea36171970293161b904b466/app/src/androidTest/java/com/pileproject/drive/execution/NxtCarControllerTest.java) for [NxtCarController](https://github.com/tiwanari/drive/blob/e53abf239ee03995ea36171970293161b904b466/app/src/nxt/java/com/pileproject/drive/execution/NxtCarController.java))

tiwanari:
Hey @amiq11 , @myusak ,

Though I know I should add more detailed changes into this PR (e.g., some Javadocs, more blocks we will use), I'm almost fixing the design of this PR. So, please look over it again and give me comments.

I want to discuss some stuff;
- What do you think about the `CarControllerBase` layer?
  - This may be too much to do here
  - There are blocks that use it by downcasting
    - Should I have to move them to a specific package something like `com.pileproject.drive.programming.visual.block.car`?
- What do you think about the `BlockPreferencesSchema`?
  - I gave up spreading them to all variants because my previous suggestion (making getter methods for thresholds) was not so good.

I may miss something. Please let me know.

app/src/nxt/java/com/pileproject/drive/execution/NxtCarController.java
```diff
@@ -0,0 +1,96 @@
...
+import static com.pileproject.drive.execution.CarControllerBase.INPUT.SOUND;
+import static com.pileproject.drive.execution.CarControllerBase.INPUT.TOUCH;
+
+public class NxtCarController extends CarControllerBase {
+    private final boolean mIsLejosFirmware;
+
+    /**
+     * binds each sensor and motor to their port
+     */
+    public NxtCarController(NxtMachine machine) {
```
mandaiy: Would it be better that the class has no dependency on the preference?
Maybe it's good to deal with it in future.

And should NxtMachine know firmware information? (perhaps TouchSensor too)
I guess so.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
...
+        public static final String BUZZER = "buzzer";
+        public static final String LED = "led";
+    }
+
+    /**
+     * Internal class to contain motor properties
+     * particularly for Android activities
+     */
+    public static final class MotorProperty {
+        public static List<String> getAllMotors() {
```
mandaiy: Please change it to a static final field something like ``List<String> ALL_MOTORS = Collections.unmodifiableList(...)``

app/src/nxt/java/com/pileproject/drive/execution/NxtCarController.java
```diff
@@ -0,0 +1,96 @@
...
+
+        return isTouched;
+    }
+
+    /**
+     * Internal class to contain sensor properties
+     * particularly for Android activities
+     */
+    public static final class SensorProperty {
+        public static List<String> getAllSensors() {
```
mandaiy: Please change it to a static final field.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
...
+    public enum MotorDir {
+        Forward, Neutral, Backward
+    }
+
+    public enum MotorKind {
+        LeftMotor, RightMotor
+    }
+
+    // all input devices
+    public class INPUT {
```
mandaiy: Because it's class, it might be better to name it with CamelCase.
And I think ``InputDevice`` is more informative.

Also if an inner class has no references to it's parent class instance, it should be `static`.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
...
+        public static final String GYRO = "gyro_sensor";
+        public static final String LINE = "line_sensor";
+        public static final String RANGEFINDER = "rangefinder";
+        public static final String REMOTE_CONTROL_RECEIVER = "remote_control_receiver";
+        public static final String SOUND = "sound_sensor";
+        public static final String TOUCH = "touch_sensor";
+    }
+
+    // all output devices
+    public class OUTPUT {
```
mandaiy: Same as the above comment

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
+/**
```
mandaiy: This class deals with `machines` but the concept of `execution` package is slightly different; that package is not for controlling machines but executing programs.
So I think it's time to move machine-related classes to a separate package.
This comment applies to `MachineController` and `MachineProvider` too.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
```diff
@@ -0,0 +1,302 @@
...
+import com.pileproject.drivecommand.machine.device.port.InputPort;
+import com.pileproject.drivecommand.machine.device.port.OutputPort;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class CarControllerBase implements MachineController {
+    public static final int INIT_MOTOR_POWER = 60;
+
+    protected MachineBase mMachine = null;
```
mandaiy: These nulls are redundant (see http://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html)

app/src/main/java/com/pileproject/drive/preferences/BlockPreferencesSchema.java
```diff
@@ -20,15 +20,13 @@
 
 /**
  * Schema for "block_preferences" table
- *
- * @author <a href="mailto:tatsuyaw0c@gmail.com">Tatsuya Iwanari</a>
- * @version 1.0 21-April-2016
  */
 @Table(name = "block_preferences")
 public class BlockPreferencesSchema {
     public class THRESHOLD {
```
mandaiy: Same as the comment of `INPUT` class

.idea/copyright/pileproject.xml
```diff
@@ -2,6 +2,6 @@
...
+    <option name="notice" value="Copyright (C) 2011-2016 PILE Project, Inc. &lt;dev@pileproject.com&gt;&#10;&#10;Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);&#10;you may not use this file except in compliance with the License.&#10;You may obtain a copy of the License at&#10;&#10;     http://www.apache.org/licenses/LICENSE-2.0&#10; &#10;Unless required by applicable law or agreed to in writing, software&#10;distributed under the License is distributed on an &quot;AS IS&quot; BASIS,&#10;WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.&#10;See the License for the specific language governing permissions and&#10;limitations under the License." />
```
mandaiy, line 5: We need 2017 soon :p

mandaiy (changes requested):
Vote for the car package of blocks.
And we use dictionary-form verbs as a first word of commit messages.

.idea/copyright/pileproject.xml
tiwanari: Yeah, that's right. Please remind me every time when a new year comes... just kidding :P 

BTW, while this is a completely different issue, I'm wondering to use the name of company `PILE Project, Inc.` is not proper way because we should care about other developers.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
tiwanari: Sounds good. Let me think where is the best.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
tiwanari: Thank for pointing out! 👍

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
tiwanari: Thanks 👍

app/src/nxt/java/com/pileproject/drive/execution/NxtCarController.java
tiwanari: > Would it be better that the class has no dependency on the preference?

So, how do we deal with it? Do you mean that we should connect devices with ports by some setter methods? I'm sorry I may not understand.


> And should NxtMachine know firmware information? 

Hmm, I'm worried about it may introduce some complexities to the library [drivecommand] because leJOS is just a variant of firmware, though I'm not so familiar wth others. Supporting only the standard firmware is not bad I think.

tiwanari (commented):
Thank you for your comments, @myusak 
I replied to your comments :)

tiwanari:
So far, I received thoughtful comments to keep the design (packages) and codes clean 👍  I'll apply them to codes :) 

Is there any other opinion about the designs? If not, I will concentrate on cleaning up codes and ask you to review again later 🚀

.idea/copyright/pileproject.xml
mandaiy: How do other companies manage it?
You might find a good way in chromium project or other open source projects.

app/src/nxt/java/com/pileproject/drive/execution/NxtCarController.java
mandaiy: > connect devices with ports by some setter methods?

That's one way to do it (but I hate setters because they introduce inconsistent states of instances).
I'm worried that this class has difficulty in testing because it directly depends on Android preference and we can't mock it.

Adding a thin interface which wraps the preference seems another way.
Or you can use the builder-pattern.

.idea/copyright/pileproject.xml
tiwanari: Oh, I forgot to show a solution I thought. For example, the Dagger project uses `The Dagger Authors` and I think this also can be applied to our products something like `The Drive Authors`.

app/src/nxt/java/com/pileproject/drive/execution/NxtCarController.java
tiwanari: Okay, I understood. That is what you said somewhere - "Reconsider to prepare `NxtControllerBuilder`" , right?

tiwanari:
Chages;

- I made `machine` package for `Controller`s and `MachineProvider`s.
- I made `*.block.(sequence|selection).car` package for `*Block`s that use downcast from `MachineController` to `CarControllerBase`.

I want to leave the `ControllerBuilder` improvement to other PR.

app/src/main/java/com/pileproject/drive/execution/CarControllerBase.java
tiwanari: I moved them into `machine` package.

tiwanari:
Hey @myusak, @amiq11 

Now, I think this PR is on the real reviewing stage.

app/src/androidTest/java/com/pileproject/drive/execution/NxtCarControllerTest.java
```diff
@@ -181,16 +186,16 @@ public void whenMotorSpeedsAreNotInitialized_thenMovesForwardWithDefaultValue()
 
         controller.moveForward();
 
-        verify(leftMotor).setSpeed(NxtController.INIT_MOTOR_POWER);
-        verify(rightMotor).setSpeed(NxtController.INIT_MOTOR_POWER);
+        verify(leftMotor).setSpeed(CarControllerBase.MotorProperty.INIT_MOTOR_POWER);
```
mandaiy, line 189: Should be ``*NxtCarController*.MoterProperty.INIT_MOTOR_POWER``? If not, why?

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+        LeftMotor, RightMotor
+    }
+
+    /**
+     * Internal class that contains motor properties.
+     */
+    public static final class MotorProperty {
+        public static final int INIT_MOTOR_POWER = 60;
+        public static final List<String> ALL_MOTORS =
+                Collections.unmodifiableList(new LinkedList<String>() {
```
mandaiy: Why not use ``Arrays.asList(OutputDevice.LEFT_MOTOER, OutputDevice.RIGHT_MOTOR)``

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+    public void turnOnLed() {
+        throw new UnsupportedOperationException("This machine does not support 'turnOnLed' command");
+    }
+
+    public void turnOffLed() {
+        throw new UnsupportedOperationException("This machine does not support 'turnOffLed' command");
+    }
+
+    /**
+     * move forward
```
mandaiy: Should be "Moves"

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+
+import static com.pileproject.drive.machine.CarControllerBase.MotorProperty.INIT_MOTOR_POWER;
+
+/**
+ * This is a base machine controller which controls a car-formed machine (e.g., Nxt/Ev3/Pile robot). A car-formed
+ * machine is a machine that at least has left and right {@link Motor}s. It depends on the machine what kinds of
```
mandaiy: I can't understand the meaning of the sentence "It depends on the machine what kinds of devices..."
What is "It" at the beginning of the sentence?

and the sentence has an unmatched parenthesis, possibly you should put next to ``{@link TouchSensor}``

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+ * devices (e.g., {@link Buzzer}, {@link TouchSensor} it actually has, and such information (e.g.,
+ * methods/properties) should be treated in child controllers like {@link NxtCarController}.
+ *
+ * This base class must specify the interface to control a car-formed machine as methods which are the union of
+ * methods of child controllers, and provide the list of all input/output devices may be used (see: {@link InputDevice},
```
mandaiy: "provide lists of all input/output devices *which* may be used"

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+ * {@link OutputDevice}).
+ * NOTE: Because of this, if you want add a new controller with a new feature which extends this base class , you
```
mandaiy: "Because of *the design of the controllers*, if you want *to* add..."

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+ * should add a new one in this base class.
```
mandaiy: "should add *the methods you defined in your new controller to* this base class"

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * Because a car-formed machine must have two motors (left and right), this base class has the properties of these
+ * motors (see: {@link CarControllerBase.MotorProperty}). However, as mentioned
+ * above, the information of other devices should have been specified in child controllers. For example, we keep the
```
mandaiy: "... devices should *be* specified..."
(should have p.p: http://ell.stackexchange.com/questions/90993/meaning-of-should-havep-p)

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
mandaiy: I would write "A car-formed machine is a machine which has two motors (left/right) and arbitrary number of sensors."
(Of course it's better to add notes that the motors have to be arranged in parallel and the directions of motor-rotation have to be same and... but it's overkill.)

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
```diff
@@ -0,0 +1,339 @@
...
+                    mRightMotor.setSpeed(mRightMotorPower);
+                    mRightMotor.backward();
+                    break;
+            }
+        }
+    }
+
+    /**
+     * A setter of motor power.
+     *  @param kind see {@link MotorKind}
```
mandaiy: unnecessary space

mandaiy (approved):
Looks good to me, though some public methods do not have javadoc and there are a few things you can do before merging.

app/src/main/java/com/pileproject/drive/machine/CarControllerBase.java
tiwanari: I intended 
It = `what kind of ~`

And thanks for pointing out the missing parenthesis.

tiwanari: Yeah, `car-formed` is not yet clear enough actually. We may have to prepare some pictures somewhere (e.g., wiki) for describing them.

tiwanari (commented):
I greatly appreciate your review. To tell the truth, I totally forgot to add JavaDocs to some methods :) Thank you for pointing out.

I will add/fix them soon!

app/src/androidTest/java/com/pileproject/drive/execution/NxtCarControllerTest.java
tiwanari: This is just because the value is defined in `CarControllerBase` not in `NxtCarController`. I thought the children will share the initial value for motors.

tiwanari:
Hey @amiq11 @myusak,

If there are no more concerns, I will merge this :)

app/src/androidTest/java/com/pileproject/drive/execution/NxtCarControllerTest.java
mandaiy: then why do you use that of nxt for the right motor in the below line?
am i seeing some old version?

tiwanari: oh i didn't see other lines. thanks

tiwanari:
I applied your comments and fix codes a little.

tiwanari:
I want to merge this around the end of tomorrow. Is it okay?
Please let me know if you have more opinions or are too busy to check.

app/src/main/java/com/pileproject/drive/programming/visual/block/BlockFactory.java
```diff
@@ -33,20 +33,17 @@
...
 public class BlockFactory {
 
     private BlockFactory() {
         throw new AssertionError("This class cannot be instantiated");
     }
 
+    @SuppressWarnings("unchecked")
     private static <T extends BlockBase> Class<T> getClassForName(String className) throws RuntimeException {
         try {
-            // noinspection unchecked
```
mandaiy, line 49: Do you hate intellij extension?
According to Effective Java, we have to prefer as small scope of ``@SuppressWarnings`` as possible.
Rather than annotating on classes or methods, we can add SuppressWarnings on this single line here.

There are two ways: use ``// noinspection unchecked`` or, add an assignment statement for a ``Class<T>`` variable and use ``@SuppressWarnings`` (and return the variable).

tiwanari: Thank you for your suggestion. I took the latter because the IDE-specific feature didn't suppress the warning like [this](https://travis-ci.org/PileProject/drive/builds/184892292#L2407) and it was annoying.

tiwanari:
Hey @amiq11,

WDYT?  Let me know your opinion.

tiwanari:
I didn't receive any objection to merging so far. So, let me go ahead :)

mandaiy:
Hey, don't skip the commit message and leave it to the github.
At least you should add a summary that describes what classes are changed and why
so that people do not have to come to this pull request.

tiwanari:
Thank you for the comment.
I didn't care about it and very sorry about that.
I added the detail and used force commit (disable the protection of `develop` branch temporary). 
I promise to keep it in mind.

Issue #56, opened 2016-11-08
Javadocs geneation with Gradle


tiwanari:
### Problem
`Drive` has some Javadocs but we don't have a good way to generate them.

### Proposal
Do Javadocs generation as a task of Gradle. It should be something like this;

```
./gradlew generateNxtJavadocs
```

tiwanari:
I'm trying to use [vanniktech/gradle-android-javadoc-plugin: Gradle plugin that generates Java Documentation from an Android Gradle project.](https://goo.gl/BozlQu) as @myusak suggested offline but I got some errors related to auto-generated classes such as android's `R` and `com.pileproject.drive.preferences.BlockPreferences` ...  :cry:
(I also get some minor errors)

tiwanari:
This is all I changed;

``` diff
diff --git app/build.gradle app/build.gradle
index 27c36b4..9b67cf6 100644
--- app/build.gradle
+++ app/build.gradle
@@ -16,6 +16,7 @@

 apply plugin: 'com.android.application'
 apply plugin: 'com.neenbedankt.android-apt'
+apply plugin: 'com.vanniktech.android.javadoc'

 android {
     compileSdkVersion 23
diff --git build.gradle build.gradle
index 53c9d7d..542ffdc 100644
--- build.gradle
+++ build.gradle
@@ -3,11 +3,15 @@
 buildscript {
     repositories {
         jcenter()
+        mavenCentral()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.1.0'

         classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
+
+        classpath 'com.vanniktech:gradle-android-javadoc-plugin:0.2.1'
+
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
```

tiwanari:
I fixed these errors on #62.

Issue #24, opened 2016-07-25
Missing comments/javadocs


tiwanari:
Please add comments/javadocs and then check files.
- [ ] ./androidTest/java/com/pileproject/drive/ApplicationTest.java
- [ ] ./main/java/com/pileproject/drive/app/DriveApplication.java
- [ ] ./main/java/com/pileproject/drive/comm/BluetoothCommunicator.java
- [ ] ./main/java/com/pileproject/drive/database/DriveDatabase.java
- [ ] ./main/java/com/pileproject/drive/database/ProgramDataManager.java
- [ ] ./main/java/com/pileproject/drive/database/ProgramDataSpec.java
- [ ] ./main/java/com/pileproject/drive/database/ProgramSpec.java
- [ ] ./main/java/com/pileproject/drive/execution/ExecutionActivityBase.java
- [ ] ./main/java/com/pileproject/drive/execution/ExecutionCondition.java
- [ ] ./main/java/com/pileproject/drive/execution/ExecutionThread.java
- [ ] ./main/java/com/pileproject/drive/execution/MachineController.java
- [ ] ./main/java/com/pileproject/drive/preferences/CommonPreferencesSchema.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/activity/BlockIconAdapter.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/activity/BlockListActivityBase.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/activity/BlockPositionComparator.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/activity/ProgramListFragment.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivityBase.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/BlockBase.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/BlockFactory.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/NumTextHolder.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/repetition/RepetitionBlock.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/repetition/RepetitionBreakBlock.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/repetition/RepetitionEndBlock.java
- [ ] ./main/java/com/pileproject/drive/programming/visual/block/repetition/RepetitionHasNumText.java
[...]

PR #62, opened 2016-12-20
Clean up comments

tiwanari:
Solves #24. This is a wip PR to clean up comments/javadocs. This may be related to #56.

# Background
Before moving on to adding other variants (ev3/pile), I think we should fix some comments/javadocs because we should do the same changes to the other variants if we implement them first. At present, some javadocs have errors with their grammar.

# Goal
The goal is not enriching but cleaning comments. Of course, I will add some information but they may be not enough. If you really want more information on specific ones, please let me know (e.g., `NxtCarController` needs a specification of its form).

# What kinds of help I need
I will fix them little by little so please give this PR any comments and correct English.


tiwanari:
NOTE: 
I will remove the plugin I just added because it doesn't work properly for auto-generated files. But it is helpful to check javadocs.

tiwanari:
To show an example of JavaDocs, I prepared [AlertDialogFragment.java](https://github.com/PileProject/drive/pull/62/files#diff-b5ade28d9d71622d1b73773ed58eb8b7) (I chose it because it had been almost finished). If you don't have enough time, firstly please check it alone and give comments.

tiwanari:
Hey @myusak and @amiq11,

Now, this PR is ready for review.

I modified comments to make them follow the rule of JavaDocs and unified the style of comments/JavaDocs. 

While I said the JavaDocs generation plugin would be removed because it emits errors/warnings related to auto-generated classes, I think it is okay to keep it as long as we avoid using the reference to the auto-generated classes. For example, we CANNOT write

```java
... {@link MachinePreference}
```

because it causes a reference error. So, we should write something like

```java
... <code>MachinePreferences</code> which is generated based on {@link MachinePreferencesSchema}
```

Now, we have the clean (no error) codes for JavaDocs generation (but have some warnings related to the auto-generated classes).

I know this PR is too huge to go into but please check them roughly.

NOTE:
- I updated the `LICENSE` according to the change of `drivecommand`'s LICENSE and applied it to all the codes with `licenseFormat` task. This will help us to make sure that I checked all the codes by showing all the classes in this PR diff screen and if you find a class that has changes only in the LICENSE header, please check if I forgot to modify other comments.
- I renamed a util class to follow the naming style of others.

tiwanari:
Please refer to the oracle's [document](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html) for javadocs.

mandaiy (approved):
Comments theirselves seem good enough to me, though
I don't know it's good or not to introduce the plugin.
I'll leave it to you

tiwanari:
@myusak Thank you for your review 👍  Yeah, it's controversial.

So, what do you think, @amiq11 ?

makotoshimazu:
Thanks for working on that!
I cannot have enough time to investigate thoroughly these days, so I'd like to leave it to you...
One thing I came up with is that "The Drive Authors" seems a bit narrow. I prefer using the same notice with other OSS products like drivecommand or upcoming hardware, so I feel it should include "PILE" or something. (I couldn't get good words, though...)

tiwanari:
@amiq11 Thank you for your kind confirmation :sparkles: Okay, I will decide it.

As for the author notice, I have a different idea that `PILE` is narrower than `The Drive Authors`. I think `PILE` refers to us but I want, yeah hopefully, others to help us develop our products. That's why I avoided `PILE Project, inc.` Maybe, `PILE Product Developers` or something like that meets your and my intention but I'm also not sure what the best is 😕 

To tell the truth, that kind of notice just came from Google's [Dagger](https://github.com/google/dagger/blob/master/LICENSE.txt) :P

makotoshimazu:
> Maybe, PILE Product Developers or something like that meets your and my intention but I'm also not sure what the best is :confused:

Yes, that's just what I've imagined.

makotoshimazu:
And "The PILE Developers" looks good for me. WDYT?

tiwanari:
looks good to me :+1:

tiwanari:
Hi there

I fixed the author notice and now I'm going to merge this PR. 

Finally, I decided to include the plugin which introduces some tasks like `generateNxtDebugJavadoc`. Running these tasks causes some warnings because of the reference errors of auto-generated classes but we can see the complete Javadocs in `app/javaDoc`. I believe that the document is really helpful not only for others but also for us.

However, I'm sorry that we should care about the tasks so that they shouldn't have errors related to auto-generated classes. But please don't worry about that. I will check them when you submit a new PR.

PR #67, opened 2017-06-06
Fix a bug on validation for MAC address

tiwanari:
I found a small but fatal bug. 

The validation of a MAC address is wrong. If the string is not set, `""` will be returned not null.

### DIFF
--- a/app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivity.java
+++ b/app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivity.java
@@ -97,7 +97,7 @@ public class ProgrammingActivity extends AppCompatActivity implements AlertDialo
         mSpaceManager.saveExecutionProgram();
 
         // TODO: remove bluetooth-dependent code for WiFiCommunicator
-        String address = MachinePreferences.get(getApplicationContext()).getMacAddress();
+        String address = MachinePreferences.get(getApplicationContext()).getMacAddress("");
 
         if (!address.equals("") || DeployUtil.isOnEmulator()) {
             Intent intent = ExecutionActivity.createIntent(getApplicationContext());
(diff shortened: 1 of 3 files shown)


mandaiy (commented):
At a glance, we have a similar validation on L138.
Doesn't the validation work correctly too?

According to the README of KVS-schema, we can specify default values in schema classes.
The README seems not to tell what the default value of string is if we don't provide a default value explicitly (and the default value is actually an empty string, right?).
So I think we should add the default value for the address explicitly (empty string seems appropriate).
Could you add the default value for the string?

tiwanari:
> Doesn't the validation work correctly too?

No, thank you for pointing out that!

Your suggestion is quite reasonable. Will do:)

tiwanari:
I did:)

mandaiy:
Ah, I meant that why not add the default value to the definition of MachinePreference like
```
@Table(name = "machine_preferences")
public class MachinePreferencesSchema {
    // ...
    @Key(name = "bluetooth_address")
    String macAddress = "";
    // ...
```

But calling ``getMacAddress()`` with providing a default argument also makes sense and might be enough.
It's up to you whether adding the default value to the class definition or not.

mandaiy (approved):

tiwanari:
Ah, sorry for my misunderstanding.. Hmm, in this case, both of them are reasonable but this kind of bugs is easily made and confusing. I guess this is a good time to start using guava which has useful methods like `isNullOrEmpty` . What do you think?

tiwanari:
I think it's natural the method returns `null` (that's why the bug was made ) and returning `""` as well (in the Java world). So, using the until method, we don't have to specify the default value from my viewpoint.

app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivity.java
```diff
@@ -134,9 +134,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
     private void setUpToolbar() {
         // get device address to show it on toolbar
-        String deviceAddress = MachinePreferences.get(getApplicationContext()).getMacAddress();
+        String deviceAddress = MachinePreferences.get(getApplicationContext()).getMacAddress("");
```
mandaiy: If I were you I'd write
```java
String defaultString = getResources().getString(R.string.programming_noTargetDevice);
String deviceAddress = MachinePreferences.get(getApplicationContext()).getMacAddress(defaultString);
```

mandaiy:
Introducing guava is fine to me.
But I think there are a lot of lines we can refactor by using the library and I want to replace such lines at a time.

So can you close this pull request?
And let's discuss the library and list up the lines in another issue or pull request.

app/src/main/java/com/pileproject/drive/programming/visual/activity/ProgrammingActivity.java
tiwanari: looks good. thanks!

tiwanari:
okay, after modifying lines you mentioned, I'll close this and raise an issue

